PR #12, opened 2018-01-23
Improve handling of user close action

crookedneighbor:
Previously, there was no clear indication that the window had been closed by the user. This now passes back a specific error to inform the developer that the window has been closed.

I think this would constitute a breaking change, but since we're still in the `0.x.x` namespace, I think it's fine to just bump to `0.2.0` for the next release.

@braintree/team-dx

### DIFF
--- a/PopupBridge/src/main/java/com/braintreepayments/popupbridge/PopupBridge.java
+++ b/PopupBridge/src/main/java/com/braintreepayments/popupbridge/PopupBridge.java
@@ -125,6 +125,8 @@ public class PopupBridge extends BrowserSwitchFragment {
             } catch (JSONException ignored) {}
 
             payload = json.toString();
+        } else if (result == BrowserSwitchResult.CANCELED) {
+            error = "new Error('POPUP_CLOSED')";
         } else if (result == BrowserSwitchResult.ERROR) {
             error = "new Error('" + result.getErrorMessage() + "')";
         }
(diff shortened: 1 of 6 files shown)


crookedneighbor:
I don't think we can actually do this. In our implementation in braintree-web we rely on not having a payload or error object to mean that the user dismissed the window. Changing this logic will break people using braintree-web's PayPal Component.

https://github.com/braintree/braintree-web/blob/f5cea4716853fdb011d55116341931304cdfdcca/src/lib/frame-service/external/strategies/popup-bridge.js#L17

crookedneighbor:
We've updated this to instead call a new `onClose` method if it exists and call `onComplete(null, null)` if it does not.

@skunkworks thoughts?

PopupBridge/src/main/java/com/braintreepayments/popupbridge/PopupBridge.java
```diff
@@ -125,6 +125,14 @@ public void onBrowserSwitchResult(int requestCode, BrowserSwitchResult result, U
             } catch (JSONException ignored) {}
 
             payload = json.toString();
+        } else if (result == BrowserSwitchResult.CANCELED) {
```
bluk: Is there a difference in behavior here between the iOS PR and this? It seems this will call `onComplete` twice or `onClose()` and `onComplete` if it exists.

crookedneighbor: How does it call it twice? It has an early return so it doesn't call it outside of the if block.

bluk: Ahh I missed the `return` statement.

PR #19, opened 2018-10-04
Returns URI fragments

quinnjn:
Adding parity for the [iOS PR](https://github.com/braintree/popup-bridge-ios/pull/13)

Question: Do we want the key to be `fragment`, or `hash`? Any other suggestions?

### DIFF
--- a/PopupBridge/src/main/java/com/braintreepayments/popupbridge/PopupBridge.java
+++ b/PopupBridge/src/main/java/com/braintreepayments/popupbridge/PopupBridge.java
@@ -123,10 +123,10 @@ public class PopupBridge extends BrowserSwitchFragment {
 
             JSONObject json = new JSONObject();
             JSONObject queryItems = new JSONObject();
-            String fragment = returnUri.getFragment();
+            String hash = returnUri.getFragment();
 
-            if (fragment == null) {
-                fragment = "";
+            if (hash == null) {
+                hash = "";
             }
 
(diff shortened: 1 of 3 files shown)


skunkworks:
I chose `hash` to match the JS URL spec, although that may be a non-goal considering that we return `queryItems` and not `query`.

PopupBridge/src/main/java/com/braintreepayments/popupbridge/PopupBridge.java
```diff
@@ -139,6 +144,7 @@ public void onBrowserSwitchResult(int requestCode, BrowserSwitchResult result, U
             try {
                 json.put("path", returnUri.getPath());
                 json.put("queryItems", queryItems);
+                json.put("fragment", fragment);
```
skunkworks: iOS is omitting the `hash` aka `fragment` key/value pair if it's not defined. Any thoughts on which way we should go?

crookedneighbor: looks like ios is calling this `hash`, right? https://github.com/braintree/popup-bridge-ios/commit/cd45320#diff-7d6a776faf444a83619c6174c700ef2bR138

quinnjn: Omitting it sounds good to me. I'll make the change.

PR #21, opened 2020-02-28
Relax activity requirement to FragmentActivity

saschpe:
There is no need to request an instance of AppCompatActivity. Among other things, this allows testing with 'androidx.fragment-testing', which only provides a FragmentActivity to fragments.


saschpe:
Danke!

PR #34, opened 2020-09-11
Update integration tests to assert that text exists

sarahkoop:
### Summary of changes

 - Update integration tests to use `assertTrue()` instead of `waitForExists()` only.
 - Remove connectedAndroidTest and wait for emulator from travis.yml

 ### Checklist

### Authors

- @sarahkoop

### DIFF
--- a/PopupBridgeExample/src/androidTest/java/com.braintreepayments.popupbridge.example.test/PopupBridgeTest.java
+++ b/PopupBridgeExample/src/androidTest/java/com.braintreepayments.popupbridge.example.test/PopupBridgeTest.java
@@ -11,9 +11,11 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 
 import static com.lukekorth.deviceautomator.AutomatorAction.click;
+import static com.lukekorth.deviceautomator.AutomatorAssertion.text;
 import static com.lukekorth.deviceautomator.DeviceAutomator.onDevice;
 import static com.lukekorth.deviceautomator.UiObjectMatcher.withContentDescription;
 import static com.lukekorth.deviceautomator.UiObjectMatcher.withText;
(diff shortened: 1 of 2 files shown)


PopupBridgeExample/src/androidTest/java/com.braintreepayments.popupbridge.example.test/PopupBridgeTest.java
```diff
@@ -46,7 +48,8 @@ public void opensPopup_whenClickingDontLikeAnyOfTheseColors_returnsCanceledSelec
         onViewWithText("Launch Popup").perform(click());
         onViewWithText("I don't like any of these colors")
                 .waitForExists(BROWSER_TIMEOUT).perform(click());
-        onViewWithText("You do not like any of these colors").waitForExists();
```
sarahkoop, line 49: This test was previously passing even though the text did not match what was on the screen (these vs. our). Converting the test to use `containsString()` caused the test to fail as expected with non-matching text, and pass with matching text on screen.

PopupBridgeExample/src/androidTest/java/com.braintreepayments.popupbridge.example.test/PopupBridgeTest.java
```diff
@@ -46,7 +47,7 @@ public void opensPopup_whenClickingDontLikeAnyOfTheseColors_returnsCanceledSelec
...
```
sarahkoop, line 49: This test was previously passing even though the text did not match what was on the screen (do vs. did and these vs. our). Converting the test to use `assertTrue()` caused the test to fail as expected with non-matching text, and pass with matching text on screen.

sshropshire (approved):

billwerges (approved):

PR #35, opened 2020-10-02
Android 11

sarahkoop:
### Summary of changes

 - Bump `targetSdkVersion` and `compileSdkVersion` to API level 30
 - Upgrade browser-switch to 1.1.2

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sshropshire 
- @sarahkoop


.travis.yml
```diff
@@ -1,18 +1,14 @@
...
+    - android-30
+    - build-tools-30.0.2
     - platform-tools
     - extra-android-support
     - extra-android-m2repository
     - extra-google-m2repository
 before_install:
-  - yes | sdkmanager "platforms;android-28"
-script: ./gradlew --info clean lint test
-after_success: ./deploy_snapshot.sh
```
sestevens, line 18: Should we delete the `deploy_snapshot` file as well?

sshropshire: Yeah that's a good call. We've been doing it pretty ad hoc lately. It makes sense to just get rid of the file.

sestevens (approved):

sestevens:
Now that we're adding support for Android 11, we can probably close all of Quinn's old PRs to add support for Android Q, right? I think there's one in each Android repo.

sshropshire:
@sestevens yup that makes sense. Might do a quick scan to make sure there isn't something there that we're missing but it'll clear up the PR backlog for sure.

scannillo (approved):

PR #37, opened 2021-02-10
Restructure PopupBridge Module

sarahkoop:
### Summary of changes

* Add `PopupBridgeClient`
* Add `PopupBridgeErrorListener`
* Breaking Changes
  * Change package from `com.braintreepayments` to `com.braintreepayments.api`
  * Remove `PopupBridge`
  * Remove `PopupBridgeActivity`

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop 
- @sshropshire


billwerges (approved):

PopupBridge/build.gradle
```diff
@@ -22,7 +22,7 @@ android {
 }
 
 dependencies {
-    api 'com.braintreepayments:browser-switch:1.1.2'
+    implementation 'com.braintreepayments.api:browser-switch:2.0.0-beta1'
```
sestevens, line 25: Maybe we should add this to the changelog as well? I assume we'll use a non-beta version when we actually release pop-up bridge, but we could add a changelog entry now so that we don't forget later.

sestevens (approved):

PopupBridge/build.gradle
sarahkoop: Good idea - added that entry

PR #38, opened 2021-02-17
Add Migration Guide

sarahkoop:
### Summary of changes

 - Add `v4` migration guide
 
 ### Checklist

### Authors

- @sarahkoop 
- @sshropshire

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@ PopupBridge is an Android library that allows WebViews to open popup windows in 
 
 ![PopupBridge Demo gif](popup-bridge-android.gif)
 
+:mega:&nbsp;&nbsp;A new major version of the SDK is available in beta. See the [v4 migration guide](v4_MIGRATION.md) for details.
+
 PopupBridge is also available for [iOS](https://github.com/braintree/popup-bridge-ios).
 
 See the [Frequently Asked Questions](#frequently-asked-questions) to learn more about PopupBridge. See [Using PayPal in a WebView](#using-paypal-in-a-webview) to use PopupBridge with PayPal.
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ See the [Frequently Asked Questions](#frequently-asked-questions) to learn more 
 Requirements
 ------------
 
-- Android SDK 19
+- Android SDK 21
 
 Installation
 ------------
--- a/v4_MIGRATION.md
+++ b/v4_MIGRATION.md
@@ -67,11 +67,11 @@ If these requirements are not met, an error will be reported to the `PopupBridge
 
 2. Use PopupBridge from the web page by writing some JavaScript:
 
-See [README](README.md) for sample code.
+  - Refer to `v3` docs in [README](README.md) for sample code.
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -7,14 +7,16 @@ PopupBridge is an Android library that allows WebViews to open popup windows in
...
+
 PopupBridge is also available for [iOS](https://github.com/braintree/popup-bridge-ios).
 
 See the [Frequently Asked Questions](#frequently-asked-questions) to learn more about PopupBridge. See [Using PayPal in a WebView](#using-paypal-in-a-webview) to use PopupBridge with PayPal.
 
 Requirements
 ------------
 
-- Android SDK 19
+- Android SDK 21
```
scannillo, line 19: Does v3 or v4 require Android SDK 21?

v4_MIGRATION.md
```diff
@@ -0,0 +1,95 @@
...
+        <data android:scheme="my-custom-url-scheme"/>
+        <category android:name="android.intent.category.DEFAULT"/>
+        <category android:name="android.intent.category.BROWSABLE"/>
+    </intent-filter>
+</activity>
+```
+
+**Note**: The scheme you define must use all lowercase letters. This is due to [scheme matching on the Android framework being case sensitive, expecting lower case](https://developer.android.com/guide/topics/manifest/data-element#scheme).
+
+If these requirements are not met, an error will be reported to the `PopupBridgeErrorListener`, if registered.
```
scannillo, line 31: I think since this information also applies to v3 (and doesn't represent a change that came in v4), we can instead move this to the README. This seems more detailed than the current instructions in the README.

scannillo: Actually - I think this stems from a difference in how we wrote the iOS migration guide and how this one is written. For [iOS](https://github.com/braintree/braintree_ios/blob/master/V5_MIGRATION.md) we kept the migration guide specific to assuming the merchant already has a working v3 integration and only listed the additional (or breaking) changes they would need to add.

README.md
sshropshire: Both actually, this is a typo we realized we missed after bumping to `minSdkVersion` 21 in version `3.0.0`.

v4_MIGRATION.md
sshropshire: Yeah we can review the iOS guide again for consistency. There is a _slight_ difference in `v4` though. In `v3`, `com.braintreepayments.popupbridge.PopupBridgeActivity` is owned by the Braintree SDK. In `v4` docs, we mention `MyPopupBridgeActivity` to signal to the merchant that they now own the deep-link destination activity.

sestevens: +1 to having the migration guide just list the required changes for someone who already has a working v3 integration

sshropshire: Updated it a bit to mention `v3`. I think it's a fairly straightforward migration but we can always enhance the docs more before GA if we see a need.

sestevens (approved):
Looks good to me!

scannillo (approved):

PR #40, opened 2021-02-24
PopupBridge 4.0.0-beta1

sarahkoop:
### Summary of changes

 -  Merge `next-major-version` into `master` for `v4` beta 

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop 
- @sshropshire


PopupBridgeExample/build.gradle
```diff
@@ -7,14 +7,18 @@ android {
     buildToolsVersion '29.0.3'
 
     defaultConfig {
-        applicationId 'com.braintreepayments.popupbridge.example'
+        applicationId 'com.braintreepayments.popupbridge.demo'
```
sestevens, line 10: This doesn't have to happen right now, but maybe it would make sense to rename this directory to `PopupBridgeDemo` instead of `PopupBridgeExample`?

sestevens (approved):

PopupBridgeExample/build.gradle
sshropshire: Yeah that makes sense. We're looking to standardize the "Demo" application name across each repo going forward.

sshropshire: We decided to just go for it. The app module is now named Demo.

demerino (approved):

PR #45, opened 2021-04-20
V4 Upgrade Browser Switch Version

sshropshire:
### Summary of changes

 - Upgrade browser-switch to 2.0.0-beta3

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop
- @sshropshire

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,7 @@ ext {
     compileSdkVersion = 30
     minSdkVersion = 21
     targetSdkVersion = 30
-    versionCode = 11
+    versionCode = 12
     versionName = version
 }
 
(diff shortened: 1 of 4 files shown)


scannillo (approved):

build.gradle
```diff
@@ -22,7 +22,7 @@ ext {
     compileSdkVersion = 30
     minSdkVersion = 21
     targetSdkVersion = 30
-    versionCode = 11
+    versionCode = 12
```
scannillo, line 25: I'm curious - what is this?

sshropshire: These are supposed to be unique so after we released `3.x` we bumped the verison code here just in case. Although this may apply to actual apps released on the PlayStore and not to SDKs.

sshropshire: Ended up bumping the version code out of habit. Similar to the build number on iOS I believe.

scannillo: Awesome, thank you

PR #51, opened 2021-08-25
Android 12 Updates v4

sarahkoop:
### Summary of changes

 - Upgrade `targetSdkVersion` and `compileSdkVersion` to API 31
 - Bump `browser-switch` version to `2.1.0`

**Note**: The issue with switching to the Venmo app has been resolved, but on return to popup-bridge, we can't see the result correctly. After testing, I think the issue is with the JS in our example site, not popup-bridge. I tested on both Android 11 and Android 12 devices and am able to return from the Venmo app switch with a success payload and a nonce - the issue occurs with running the success javascript in the webview. This behavior is happening on the master branch too, so I don't think there is any issue related to this PR. 

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## unreleased
 
 * Upgrade `targetSdkVersion` and `compileSdkVersion` to API 31
-* Bump `browser-switch` version to `2.X.X`
+* Bump `browser-switch` version to `2.1.0`
 
 ## 4.0.1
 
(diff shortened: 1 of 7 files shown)


sarahkoop:
After further investigation, the issue returning from Venmo isn't present in v3, and it seems to be related to the `PopupActivity` in the demo app being `singleTop`, and getting re-created after returning from Venmo app switch (then the URL to load in the webview is null and the page is blank) - switching it to `singleInstance` in testing resolved and showed the Venmo result.

scannillo (approved):

PR #52, opened 2021-08-25
Android 12 Updates v3

sarahkoop:
### Summary of changes

 - Upgrade `targetSdkVersion` and `compileSdkVersion` to API 31
 - Bump `browser-switch` version to `2.1.0`

 ### Checklist

 - [x] Added a changelog entry

-@sarahkoop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # PopupBridge Android Release Notes
 
+## unreleased
+
+* Upgrade `targetSdkVersion` and `compileSdkVersion` to API 31
+* Bump `browser-switch` version to `2.1.0`
+
 ## 3.1.1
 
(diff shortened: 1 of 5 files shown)


sshropshire (approved):
🌊

scannillo:
Should v3 be getting these changes?

Also side note - this repo doesn't have a [`Versions`](https://github.com/braintree/braintree_android#versions) section in its README, which we should add.

sarahkoop:
@scannillo - We made the Android 12 updates for the current and previous major version of all the SDKs. Since popup-bridge v4 was released this year, v3 is still supported so we are making the updates. 

Agreed a versions section would be good. We don't have one for drop-in either. I'll add those in.

scannillo (approved):

PR #59, opened 2022-05-02
Add Local Payment Methods to Demo App

jaxdesmarais:
### Summary of changes

 - Add Local Payment Methods example to the demo app

 ### Checklist

- @sarahkoop @jaxdesmarais

### DIFF
--- a/Demo/src/main/AndroidManifest.xml
+++ b/Demo/src/main/AndroidManifest.xml
@@ -17,7 +17,7 @@
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
-                <data android:scheme="my-custom-url-scheme" />
+                <data android:scheme="com.braintreepayments.popupbridgeexample" />
             </intent-filter>
         </activity>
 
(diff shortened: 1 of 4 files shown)


Demo/src/main/AndroidManifest.xml
```diff
@@ -17,7 +17,7 @@
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
-                <data android:scheme="my-custom-url-scheme" />
+                <data android:scheme="com.braintreepayments.popupbridgeexample" />
```
jaxdesmarais, line 20: Updated this to the same URL scheme as iOS so the example works for both demos

sshropshire (approved):

Demo/src/main/java/com/braintreepayments/popupbridge/demo/PopupActivity.java
```diff
@@ -20,7 +20,7 @@ protected void onCreate(Bundle savedInstanceState) {
         setContentView(R.layout.activity_popup);
         mWebView = findViewById(R.id.web_view);
 
-        mPopupBridgeClient = new PopupBridgeClient(this, mWebView, "my-custom-url-scheme");
+        mPopupBridgeClient = new PopupBridgeClient(this, mWebView, "com.braintreepayments.popupbridgeexample");
```
jaxdesmarais, line 23: Updated this to the same URL scheme as iOS so the example works for both demos

andrespelaezp (approved):
Looking good!

PR #63, opened 2022-12-13
Popup Bridge Observer

sshropshire:
### Summary of changes

 - This PR creates an internal `PopupBridgeObserver` to simplify Popup Bridge integrations

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # PopupBridge Android Release Notes
 
+## unreleased
+
+* Breaking Changes
+  * Remove `PopupBridgeClient#deliverPopupBridgeResult(FragmentActivity)`
+  * Make `PopupBridgeClient#POPUP_BRIDGE_NAME` package-private
+  * Make `PopupBridgeClient#POPUP_BRIDGE_URL_HOST` package-private
+
 ## 4.1.0
 
 * Upgrade `targetSdkVersion` and `compileSdkVersion` to API 31
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## unreleased
 
+* Upgrade `targetSdkVersion` and `compileSdkVersion` to API 33
+* Disable Jetifier
 * Breaking Changes
   * Remove `PopupBridgeClient#deliverPopupBridgeResult(FragmentActivity)`
   * Make `PopupBridgeClient#POPUP_BRIDGE_NAME` package-private
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## unreleased
 
-* Upgrade `targetSdkVersion` and `compileSdkVersion` to API 33
+* Android 13 Support
+  * Upgrade `targetSdkVersion` and `compileSdkVersion` to API 33
 * Disable Jetifier
 * Breaking Changes
   * Remove `PopupBridgeClient#deliverPopupBridgeResult(FragmentActivity)`
--- a/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.java
+++ b/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.java
@@ -17,7 +17,7 @@ import java.util.Set;
 
 public class PopupBridgeClient {
 
-    // NEXT MAJOR VERSION: consider using a `com.braintreepayments...` prefixed request key
+    // TODO: consider using a `com.braintreepayments...` prefixed request key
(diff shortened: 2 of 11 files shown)


PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeLifecycleObserver.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @VisibleForTesting
+    final PopupBridgeClient popupBridgeClient;
+
+    PopupBridgeLifecycleObserver(PopupBridgeClient popupBridgeClient) {
+        this.popupBridgeClient = popupBridgeClient;
+    }
+
+    @Override
+    public void onStateChanged(@NonNull LifecycleOwner lifecycleOwner, @NonNull Lifecycle.Event event) {
```
scannillo, line 27: Will we add docstrings / update the README with how merchants should use this?

jcnoriega: @scannillo in `java` the default access modifier `package-private` so `PopupBridgeLifeCycleObserver` is only visible to clases that are on the same package. This class is an internal class to observe the context lifecycle so that the `PopupBridgeClient` can retrieve a `BrowserSwitchResult` effectevily

jcnoriega (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # PopupBridge Android Release Notes
 
+## unreleased
+
+* Upgrade `targetSdkVersion` and `compileSdkVersion` to API 33
```
jaxdesmarais, line 5:
```suggestion
* Android 13 Support
  * Upgrade `targetSdkVersion` and `compileSdkVersion` to API 33
```

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.java
```diff
@@ -17,8 +17,12 @@
 
 public class PopupBridgeClient {
 
-    public static final String POPUP_BRIDGE_NAME = "popupBridge";
-    public static final String POPUP_BRIDGE_URL_HOST = "popupbridgev1";
+    // TODO: consider using a `com.braintreepayments...` prefixed request key
```
jaxdesmarais, line 20: General question, is this TODO for this PR or for a future task?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeLifecycleObserver.java
scannillo: Ah, thats for clarifying. I think the `public` on line 27 threw me off. Should that be changed?

jcnoriega: No problem! 
No, because that is the signature of method in `LifecycleEventObserver` interface, that `PopupBridgeLifecycleObserver` implements.
More context of what is going on:
Because this does a browser switch, we need to know when the Merchants activity is back at the front, in order to do so, we register an observer to its lifecycle. Then Android informs our observer of any state change in the activity through this method. In line 28 we want to know if the state is `ON_RESUME`, meaning if the activity is at the front of the app.

sshropshire: Here it actually does need to be public since it's an `@Override`. Like Jose mentioned though the class will be hidden from merchants so they won't have access to it.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.java
sshropshire: Yeah I had `NEXT_MAJOR_VERSION` initially but it may be possible with out a major version bump. I replaced with `TODO: consider` but to be real it probably won't happen until Android `v5`.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeLifecycleObserver.java
scannillo: Awesome, thank you both for the explanation!

scannillo (approved):

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.java
sshropshire: Ended up reverting the next major version message.

PR #71, opened 2025-03-10
Make changes to update dependencies and compile the project

saperi22:
### Summary of changes

Since the project has not been updated in a while, we are doing some prep work for it to be able to work properly.
 - Update the minSdk version to 23, targetSdk to 35.
 - Update Gradle, AGP, Kotlin versions. Requires **Android Studio Meerkat** to work properly.
 - Added network security config to trust user certs in debug builds.
 - Moved to using TOML version catalogs for dependencies.
 - Updated most dependencies to their respective latest versions.
 - Converted Java files to Kotlin.

 ### Checklist

 - [ ] Added a changelog entry

- @tdchow 
- @saperi22


.github/workflows/tests.yml
```diff
@@ -6,11 +6,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repository
-        uses: actions/checkout@v2
-      - name: Set up Java 11
-        uses: actions/setup-java@v3
+        uses: actions/checkout@v4
+      - name: Set up Java 17
+        uses: actions/setup-java@v4
```
sarahkoop, line 11: Is this already updated in the release workflow? Otherwise should we extract setup java into it's own action to use across all?

Demo/src/main/AndroidManifest.xml
```diff
@@ -7,6 +7,7 @@
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:supportsRtl="true"
+        android:networkSecurityConfig="@xml/network_security_config"
```
sarahkoop, line 10: What is this needed for?

saperi22: See [this](https://developer.android.com/privacy-and-security/security-config) doc on what it is used for in general.
In our case, when using emulators on corp. network, we'd want to trust the certs we install on the device for network to not error out.

.github/workflows/tests.yml
saperi22: Good point. I'll extract it out.

saperi22: Also, no, release workflow hasn't been touched yet.

sarahkoop (approved):
👏

richherrera (approved):
🦾 🎖️

PR #72, opened 2025-03-11
Rename default branch to main

tdchow:
### Summary of changes

 - Update Rakefile after renaming default branch to main
 - Remove Travis CI status in readme

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -85,7 +85,7 @@ task :release_beta => :unit_tests do
   puts "\nDone. Commits and tags have been created. If everything appears to be in order, hit ENTER to push."
   $stdin.gets
 
-  sh "git push origin master #{version}"
+  sh "git push origin main #{version}"
 end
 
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -85,7 +85,7 @@ task :release_beta => :unit_tests do
   puts "\nDone. Commits and tags have been created. If everything appears to be in order, hit ENTER to push."
   $stdin.gets
 
-  sh "git push origin master #{version}"
+  sh "git push origin main #{version}"
```
tdchow, line 88: I don't think we're using the Rakefile for deployments anymore. Can we remove it entirely if GitHub Actions are working?

saperi22 (approved):

Rakefile
tdchow: Maybe we keep the Rakefile until we have a successful deployment using GitHub Actions.

PR #73, opened 2025-04-04
Update BrowserSwitch Version

tdchow:
### Summary of changes

 - Update BrowserSwitch to 3.0.0
 - Removed `PopupBridgeLifecycleObserver`
 - Add Mockk
 - Convert and rewrite PopupBridgeClientUnitTests

 ### Checklist

 - [ ] Added a changelog entry

-


PopupBridge/src/main/java/com/braintreepayments/api/PendingRequestRepository.kt
```diff
@@ -0,0 +1,22 @@
+package com.braintreepayments.api
+
+/**
+ * Repository responsible for storing and retrieving the pending request.
+ */
+class PendingRequestRepository {
```
tdchow, line 6: This repository will be implemented in a follow up story/PR.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -46,92 +51,90 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") @VisibleForTesting
...
-        })
+    fun handleReturnToApp(intent: Intent) {
+        val pendingRequest = pendingRequestRepository.getPendingRequest()
+        if (pendingRequest == null) return
+
+        pendingRequestRepository.clearPendingRequest()
+
+        when (val browserSwitchFinalResult = browserSwitchClient.completeRequest(intent, pendingRequest)) {
+            is BrowserSwitchFinalResult.Success -> runSuccessJavaScript(browserSwitchFinalResult.returnUrl)
+            is BrowserSwitchFinalResult.Failure -> runCanceledJavaScript()
```
tdchow, line 64: Should we run the "cancel" javascript when `completeRequest()` returns a Failure? It doesn't look like we previously had a failure case.

jaxdesmarais: 🤔 that's a good question. We don't do anything with the [error returned from ASWeb on iOS currently](https://github.com/braintree/popup-bridge-ios/blob/0aa9ce097a79ba5099b19c65bdd8ca5a6fd64b82/Sources/PopupBridge/POPPopupBridge.swift#L147). Not sure if we want to just log the error to the console or something similar? Wondering if @richherrera has thoughts.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -46,92 +51,90 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") @VisibleForTesting
...
-                for (queryParam in queryParams) {
-                    try {
-                        queryItems.put(queryParam, returnUri.getQueryParameter(queryParam))
-                    } catch (e: JSONException) {
-                        error = "new Error('Failed to parse query items from return URL. " +
-                                e.localizedMessage + "')"
-                    }
-                }
-            }
+        val payLoadJson = JSONObject()
```
jaxdesmarais, line 74: Super nit but should this be cased as `payloadJson` instead? That's typically how I've seen payload within our codebases.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
sarahkoop: If we have a failure scenario for both platforms maybe we should add `runFailureJavaScript()` to both?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -46,92 +51,90 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") @VisibleForTesting
...
+        when (val browserSwitchFinalResult = browserSwitchClient.completeRequest(intent, pendingRequest)) {
+            is BrowserSwitchFinalResult.Success -> runSuccessJavaScript(browserSwitchFinalResult.returnUrl)
+            is BrowserSwitchFinalResult.Failure -> runCanceledJavaScript()
+            is BrowserSwitchFinalResult.NoResult -> runCanceledJavaScript()
+        }
     }
 
-    fun onBrowserSwitchResult(result: BrowserSwitchResult) {
+    private fun runSuccessJavaScript(returnUri: Uri) {
+        if (returnUri.host != POPUP_BRIDGE_URL_HOST) return
```
sarahkoop, line 70: Should we check this before running the success scenario? What does it mean if these hosts don't match up?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -181,9 +188,9 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") @VisibleForTesting
     companion object {
         // NEXT MAJOR VERSION: consider using a `com.braintreepayments...` prefixed request key
         // to prevent shared preferences collisions with other braintree libs that use browser switch
```
sarahkoop, line 190: Is this still applicable?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
tdchow: It might be worth adding error handling for the failure case so that the merchant can notify the user. I don't think we have any javascript code for this scenario. Would we be able to reach out to someone on the web side to help us with this?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -1,21 +1,22 @@
...
 import android.webkit.WebView
-import androidx.annotation.VisibleForTesting
 import androidx.core.net.toUri
 import androidx.fragment.app.FragmentActivity
 import org.json.JSONException
 import org.json.JSONObject
 import java.lang.ref.WeakReference
 
-class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") @VisibleForTesting internal constructor(
+class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal constructor(
```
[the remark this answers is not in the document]
tdchow: I'm on the fence, but I'm open to renaming the class. I think launcher is descriptive of what the class does. 

Would it be confusing that only a Launcher class is used in the popup bridge integration? I also think there some benefit in keeping the `Client` naming convention for the main class the merchant needs to instantiate.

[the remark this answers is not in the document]
sarahkoop: Yeah I could go both ways too. Open to whatever other reviewers think!

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
tdchow: This check happens before we run the success javascript. Was there another spot where you were thinking of adding the check?

This was existing behavior, so I'm not exactly sure in what scenario the hosts would be different. Looking into the check the browser switch does, I don't think this case would ever happen. I believe if the hosts were different, `browserSwitchClient.completeRequest()` would not return `BrowserSwitchFinalResult.Success`.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
tdchow: Hmm I don't think it still applies, but I don't think I have full context on this. Do you know if SharedPreferences use to be used within browser switch to store the pending intent?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
sarahkoop: Since this will likely be a longer task (if possible) let's just mark a TODO to follow up, no need to block this PR on it

sarahkoop (approved):

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
sarahkoop: Ah yes I think it used to be - might no longe be relevant

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
tdchow: I created an additional story to handle this failure case.

agedd (approved):

PR #75, opened 2025-04-10
Add networking classes required for sending the FPTI post call

tdchow:
### Summary of changes

 - Added networking classes: `PostRequestExecutor`, `ResponseParser`, `TLSCertificate`, and `TLSSocketFactory`
     - The logic in `ResponseParser`, `TLSCertificate`, and `TLSSocketFactory` was copied from the core BT SDK
     - `PostRequestExecutor` was built ground up, only to handle the POST for the FPTI call. We can make changes to the class if additional network calls are needed.
 - Added the `kotlin-test` and `kotlinx-coroutines-test` libraries for testing coroutines

I did a quick test the classes with a sample FPTI request and received a 200 response:
Screenshot 2025-04-08 at 3 54 24 PM

 ### Checklist

~- [ ] Added a changelog entry~

- @tdchow


saperi22 (approved):

PopupBridge/src/main/java/com/braintreepayments/api/network/TLSSocketFactory.kt
```diff
@@ -0,0 +1,119 @@
...
+import javax.net.ssl.TrustManagerFactory
+
+/**
+ * `TLSSocketFactory` is a custom implementation of [SSLSocketFactory] that enforces the use of specific TLS protocols
+ * (TLSv1.2 and TLSv1.3) and allows for the management of trusted certificates from a provided [InputStream].
+ */
+internal class TLSSocketFactory(
+    certificateStream: InputStream,
+    keyStore: KeyStore = KeyStore.getInstance(KeyStore.getDefaultType()),
+    certificateFactory: CertificateFactory = CertificateFactory.getInstance("X.509"),
```
sarahkoop, line 23: Did we confirm that we need the certificates to hit the FPTI API? I wasn't sure if those were BT API specific

tdchow: The BT SDK is setting the same `SSLSocketFactory` on the `HttpsURLConnection` instance, even for the FPTI API call. I don't believe the API is requiring the cert, but the cert pinning would happen within the Android app. If the certificate didn't match - the connection would be blocked on the app side. At least that is my understanding of cert pinning 😅

sarahkoop (approved):

PopupBridge/src/main/java/com/braintreepayments/api/network/TLSSocketFactory.kt
saperi22: This is true. Rich showed me how to test cert pinning works and I got to test it out for the certificates we are extending.

jaxdesmarais: I don't think we need certs for this API but will let @richherrera confirm. If we don't need them we should avoid pinning them to avoid merchants needing to make updates.

tdchow: The API is not requiring the cert as certificate pinning is entirely an app side enforcement. Is this something we don't want to have for the popup bridge SDK?

jaxdesmarais: I don't think we do for analytics. If we added something that was transporting data that needed to be secure like we do with the APIs on BT (credit card/secure payment data) we would want to. But this API is not transporting sensitive data so we should be good without pinning the certs.

PR #77, opened 2025-04-15
Add venmo app installed check

saperi22:
### Summary of changes

 - Add interface to query if venmo is installed.

 ### Checklist

 - [x] Added a changelog entry

- @saperi22

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
   * Make `PopupBridgeClient#POPUP_BRIDGE_NAME` package-private
   * Make `PopupBridgeClient#POPUP_BRIDGE_URL_HOST` package-private
   * Remove `PopupBridgeLifecycleObserver`
-* Add JS interface for Venmo app installed check
+* Add validation to check if the Venmo app is installed on the device.
 
 ## 4.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   * Make `PopupBridgeClient#POPUP_BRIDGE_NAME` package-private
   * Make `PopupBridgeClient#POPUP_BRIDGE_URL_HOST` package-private
   * Remove `PopupBridgeLifecycleObserver`
+* Add JS interface for Venmo app installed check
 
 ## 4.1.0
 
--- a/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
+++ b/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
@@ -7,6 +7,7 @@ import android.webkit.JavascriptInterface
 import android.webkit.WebView
 import androidx.core.net.toUri
 import androidx.fragment.app.FragmentActivity
+import com.braintreepayments.api.internal.isVenmoInstalled
 import org.json.JSONException
 import org.json.JSONObject
 import java.lang.ref.WeakReference
@@ -21,6 +22,7 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
     private var navigationListener: PopupBridgeNavigationListener? = null
     private var messageListener: PopupBridgeMessageListener? = null
     private var errorListener: PopupBridgeErrorListener? = null
+    private var venmoInstalled: Boolean = false
 
     /**
      * Create a new instance of [PopupBridgeClient].
@@ -48,6 +50,8 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
 
         val webView = webViewRef.get()
         requireNotNull(webView) { "WebView is null" }
+
+        venmoInstalled = activity.isVenmoInstalled()
 
         webView.settings.javaScriptEnabled = true
         webView.addJavascriptInterface(this, POPUP_BRIDGE_NAME)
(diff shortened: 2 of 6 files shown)


PopupBridge/src/main/java/com/braintreepayments/api/internal/AppInstalledChecks.kt
```diff
@@ -0,0 +1,10 @@
+package com.braintreepayments.api.internal
+
+import android.content.Context
+
+private const val PAYPAL_APP_PACKAGE = "com.paypal.android.p2pmobile"
+private const val VENMO_APP_PACKAGE = "com.venmo"
+
+fun Context.isVenmoInstalled(): Boolean = AppHelper().isAppInstalled(this, VENMO_APP_PACKAGE)
+
+fun Context.isPayPalInstalled(): Boolean = AppHelper().isAppInstalled(this, PAYPAL_APP_PACKAGE)
```
sarahkoop, line 10: We don't currently need the PayPal app installed check. We could leave it as forward thinking for potential PayPal app switch, or limit to actual needed methods currently

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
   * Make `PopupBridgeClient#POPUP_BRIDGE_NAME` package-private
   * Make `PopupBridgeClient#POPUP_BRIDGE_URL_HOST` package-private
   * Remove `PopupBridgeLifecycleObserver`
+* Add JS interface for Venmo app installed check
```
jaxdesmarais, line 14: Not sure if we want to match the changelog entries here: https://github.com/braintree/popup-bridge-ios/blob/main/CHANGELOG.md#300-2025-04-01

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppInstalledChecks.kt
jaxdesmarais: ➕ 1 not sure if we want to leave this in now and not use it or just stick to what's needed for now. We aren't yet sure of the timeline for app switch for PayPal on the BT or PPCP SDKs so may not be needed for a while.

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppHelper.kt
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    fun isAppInstalled(context: Context, packageName: String): Boolean {
+        val packageManager: PackageManager = context.packageManager
+        try {
+            packageManager.getApplicationInfo(packageName, NO_FLAGS)
+            return true;
+        } catch (_: PackageManager.NameNotFoundException) {
+            return false;
+        }
```
tdchow, line 21:
```suggestion
        return try {
            packageManager.getApplicationInfo(packageName, NO_FLAGS)
            true
        } catch (_: PackageManager.NameNotFoundException) {
            false
        }
```

Nits - we can lift the return out of the try catch and remove the semicolons.

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppHelper.kt
```diff
@@ -0,0 +1,23 @@
...
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+internal class AppHelper {
+
+    companion object {
+        private const val NO_FLAGS = 0
+    }
+
+    fun isAppInstalled(context: Context, packageName: String): Boolean {
```
tdchow, line 14: Are we able to add some test coverage for this function?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -49,6 +51,8 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
         val webView = webViewRef.get()
         requireNotNull(webView) { "WebView is null" }
 
+        venmoInstalled = activity.isVenmoInstalled()
```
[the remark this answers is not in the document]
saperi22: not really. this way the check happens only once.
do you have a preference?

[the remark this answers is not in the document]
saperi22: Actually, it looks like I did it so that I wouldn't have to keep a reference to the context to use it when the javascript interface is invoked.

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppInstalledChecks.kt
saperi22: removed

jaxdesmarais (approved):

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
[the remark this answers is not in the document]
tdchow: Ah I see. I was seeing if there was a way to not have to keep `venmoInstalled` state in the Client. But I think having a the class hold a boolean instead of Context is better 👍

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -21,6 +22,7 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
     private var navigationListener: PopupBridgeNavigationListener? = null
     private var messageListener: PopupBridgeMessageListener? = null
     private var errorListener: PopupBridgeErrorListener? = null
+    private var venmoInstalled: Boolean = false
```
tdchow, line 25: Nit - we should be able to make this a val here since it's set in the init.

```suggestion
    private val venmoInstalled: Boolean = false
```

tdchow (approved):
Looks good! just had a small nit.

PR #79, opened 2025-04-28
Implement PendingRequestRepository

tdchow:
### Summary of changes

 - Add the DataStore androidx library
 - Implement the functions in `PendingRequestRepository` using DataStore
 - Add `CoroutineTestRule` for easier testing of coroutines
 - Add `DataStoreMock` which is an in memory DataStore used for testing

 ### Checklist

 - [ ] Added a changelog entry

-


PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -6,25 +6,45 @@ import android.net.Uri
...
+import kotlinx.coroutines.launch
 import org.json.JSONException
 import org.json.JSONObject
 import java.lang.ref.WeakReference
 
 class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal constructor(
-    private val activityRef: WeakReference<FragmentActivity>,
-    private val webViewRef: WeakReference<WebView>,
+    activity: FragmentActivity,
+    webView: WebView,
```
tdchow, line 22: I moved the WeakReferences out of the constructor to make it easier to reference `activity` in the default values of the constructor. WeakReference can return a null value which needed to be handled when referencing the activity.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -77,15 +98,24 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
...
-        val pendingRequest = pendingRequestRepository.getPendingRequest()
-        if (pendingRequest == null) return
+        if (isHandlingReturnToApp) return
+        isHandlingReturnToApp = true
+
+        coroutineScope.launch {
+            val pendingRequest = pendingRequestRepository.getPendingRequest()
+            if (pendingRequest == null) {
+                isHandlingReturnToApp = false
+                return@launch
```
sarahkoop, line 108: This differs from the core SDK where we now have merchants handle the pending request themselves to give them flexibility in the user cancelled scenario. Did we decide or confirm that we are ok with the "did not complete flow / navigated away" scenario is returned as a cancellation and can't be recovered for popup bridge? Or does that scenario not exist here?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -6,25 +6,45 @@ import android.net.Uri
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
 import org.json.JSONException
 import org.json.JSONObject
 import java.lang.ref.WeakReference
 
 class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal constructor(
-    private val activityRef: WeakReference<FragmentActivity>,
-    private val webViewRef: WeakReference<WebView>,
+    activity: FragmentActivity,
```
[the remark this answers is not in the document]
tdchow: Good catch. This was missed. I'll create a follow up story for updating this to `ComponentActivity`.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
tdchow: I think Popup Bridge slightly differs from the core SDK, where if the user navigates away from the CCT and somehow returns to the merchant's app, the WebView would be displayed. The merchant would then need to somehow notify their WebView to display messaging to return to the browser to complete the flow.

Having the merchant save the pending request would require them to implement this callback, since the web page would be launching the browser flow: 
```kotlin
fun interface PopupBridgeNavigationListener {
    fun onUrlOpened(url: String?)
}
```

I did some testing in the demo app, and it seems that the CCT appears within the demo app in the multi tasking view. Are there steps to reproduce the "did not complete flow / navigated away" scenario?

Happy to revisit this or continue discussing if we feel that we need to support this.

sarahkoop (approved):

saperi22 (approved):

PR #80, opened 2025-04-30
Add AnalyticsClient and supporting classes

tdchow:
### Summary of changes

 - Add `AnalyticsClient` class
 - Add `AnalyticsApi` which contains the FPTI api call which includes assembling the json request
 - Add `AnalyticsParamRepository` which handles the lifecycle of the sessionId value
 - Add `DeviceRepository` which handles the fetching of device data
 - Add `Time` - wrapper class around `System.currentTimeMillis()`
 - Add `UUIDHelper` for generating UUID values

 ### Checklist

 - [ ] Added a changelog entry

-


saperi22 (approved):

sarahkoop (approved):

PopupBridge/src/main/java/com/braintreepayments/api/internal/AnalyticsApi.kt
```diff
@@ -0,0 +1,84 @@
...
+            })
+        }
+
+        return JSONObject().apply {
+            put(KEY_EVENTS, events)
+        }
+    }
+
+    companion object {
+        private const val FPTI_URL = "https://api-m.paypal.com/v1/tracking/batch/events"
```
jaxdesmarais, line 62: On iOS we removed to `-m` a while ago (this PR: https://github.com/braintree/braintree_ios/pull/1175) not sure why we didn't align that on Android but this may be a good opportunity to do that.

tdchow: Thanks for calling this out! I removed the `-m` and retested the API call.

jaxdesmarais (approved):

PR #83, opened 2025-05-09
Notify caller of browser switch failure

saperi22:

### Summary of changes

 - Adding a new method to handle browser switch failure case.
 - In case of a browser switch failure, the error from `BrowserSwitchFinalResult.Failure` is wrapped in an `Error` object and sent to `onComplete`.

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
+++ b/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
@@ -124,7 +124,7 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
 
             when (val browserSwitchFinalResult = browserSwitchClient.completeRequest(intent, pendingRequest)) {
                 is BrowserSwitchFinalResult.Success -> runNotifyCompleteJavaScript(browserSwitchFinalResult.returnUrl)
-                is BrowserSwitchFinalResult.Failure -> runCanceledJavaScript()
+                is BrowserSwitchFinalResult.Failure -> runFailureJavaScript(browserSwitchFinalResult.error)
                 is BrowserSwitchFinalResult.NoResult -> runCanceledJavaScript()
             }
             isHandlingReturnToApp = false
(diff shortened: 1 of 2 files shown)


PopupBridge/src/test/java/com/braintreepayments/api/PopupBridgeClientUnitTest.kt
```diff
@@ -226,19 +226,22 @@ class PopupBridgeClientUnitTest {
     fun `when handleReturnToApp is called and browser switch fails, canceled javascript is run`() = runTest {
```
tdchow, line 226: Can we update the test name to call out the error instead of the canceled javascript?

PopupBridge/src/test/java/com/braintreepayments/api/PopupBridgeClientUnitTest.kt
```diff
@@ -226,19 +226,22 @@ class PopupBridgeClientUnitTest {
...
         verify {
-            webViewMock.evaluateJavascript(
-                withArg { javascriptString ->
-                    assertEquals(CANCELED_JAVASCRIPT, javascriptString)
-                }, null
-            )
+            webViewMock.evaluateJavascript(withArg { javascriptString ->
+                    assertTrue(
+                        javascriptString.contains(
+                            """new Error('${exception.message}')"""
```
tdchow, line 241: Do we want to verify the entire javascript snippet here?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -124,7 +124,7 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
 
             when (val browserSwitchFinalResult = browserSwitchClient.completeRequest(intent, pendingRequest)) {
                 is BrowserSwitchFinalResult.Success -> runNotifyCompleteJavaScript(browserSwitchFinalResult.returnUrl)
-                is BrowserSwitchFinalResult.Failure -> runCanceledJavaScript()
+                is BrowserSwitchFinalResult.Failure -> runFailureJavaScript(browserSwitchFinalResult.error)
```
jaxdesmarais, line 127: Maybe a silly question, but do we need a CHANGELOG entry for this at all? Not sure how/if it changes the public interface.

jaxdesmarais (approved):

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
saperi22: I think this is more of an internal change -- I'd assume the JS side of it is already documented?

tdchow (approved):

PR #84, opened 2025-05-20
update plugin

saperi22:
### Summary of changes

 - Update publishing

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/PopupBridge/build.gradle
+++ b/PopupBridge/build.gradle
@@ -51,69 +51,12 @@ dependencies {
 
 // region signing and publishing
 
-task javadocs(type: Javadoc) {
-    source = android.sourceSets.main.java.srcDirs
-    classpath += project.files(android.getBootClasspath().join(File.pathSeparator))
-    failOnError false
-}
+project.ext.name = "popup-bridge"
+project.ext.pom_name = "popup-bridge"
+project.ext.group_id = "com.braintreepayments.api"
+project.ext.version = rootProject.version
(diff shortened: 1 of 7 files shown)


PopupBridge/build.gradle
```diff
@@ -51,69 +51,12 @@ dependencies {
 
 // region signing and publishing
 
-task javadocs(type: Javadoc) {
```
tdchow, line 54: Do we need to bring in Dokka for generating/publishing reference docs?

tdchow (approved):
Changes look good! Is there any way to test the publishing changes without actually publishing to maven central?

saperi22:
> Changes look good! Is there any way to test the publishing changes without actually publishing to maven central?

I am not sure. I'm hoping to publish using snapshot and check for the validity of the changes.

PopupBridge/build.gradle
jaxdesmarais: I don't think we are publishing any reference docs for iOS currently. Do we think this is big enough that it's necessary or should dev docs eventually be fine for anything public?

jaxdesmarais (approved):

PopupBridge/build.gradle
tdchow: Ah I didn't know we didn't have reference docs for Popup Bridge. I'm good omitting them if we didn't have them before.

PR #85, opened 2025-05-21
Set isVenmoInstalled on window.popupBridge on client init

saperi22:

### Summary of changes

 - Instead of exposing a `@JavascriptInterface` that the web view needs to call, I'm directly setting `isVenmoInstalled` onto `window.popupBridge` directly. This is how iOS does it today and reduce overhead on how it is done on web.

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
+++ b/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
@@ -89,12 +89,12 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
         webView.addJavascriptInterface(popupBridgeJavascriptInterface, POPUP_BRIDGE_NAME)
 
         with(popupBridgeJavascriptInterface) {
-            venmoInstalled = activity.isVenmoInstalled()
             onOpen = { url -> openUrl(url) }
             onSendMessage = { messageName, data ->
                 messageListener?.onMessageReceived(messageName, data)
             }
         }
+        setVenmoInstalled(activity.isVenmoInstalled())
     }
 
     fun setNavigationListener(listener: PopupBridgeNavigationListener?) {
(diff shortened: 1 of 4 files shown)


tdchow (approved):

jaxdesmarais (approved):

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -87,9 +88,14 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
 
         webView.settings.javaScriptEnabled = true
         webView.addJavascriptInterface(popupBridgeJavascriptInterface, POPUP_BRIDGE_NAME)
+        webView.webViewClient = object : WebViewClient() {
```
saperi22, line 91: Calling `setVenmoInstalled(activity.isVenmoInstalled())` directly on init{} block wasn't working. I had to wait until the page loaded before setting the value.

PopupBridge/src/test/java/com/braintreepayments/api/PopupBridgeClientUnitTest.kt
```diff
@@ -65,7 +66,7 @@ class PopupBridgeClientUnitTest {
         webView: WebView = webViewMock,
         additionalMocks: () -> Unit = {}
     ) {
-        every { webViewMock.post(capture(runnableSlot)) } returns true
+        every { webView.post(capture(runnableSlot)) } returns true
```
saperi22, line 69: @tdchow I changed this to use the local `webView` object instead of the mock. In the tests I've added, I changed to using spy on webview.
Do you foresee webview ever being used as an actual webview object? In that case, the every block would fail. Should I keep this the way it is now or do you prefer if I had an every block locally within the test function to stub out the behaviour?

jaxdesmarais (approved):

PopupBridge/src/test/java/com/braintreepayments/api/PopupBridgeClientUnitTest.kt
tdchow: Sorry I missed this comment! I think using the local webView is okay here. We can always update the tests if need later down the line.

PR #86, opened 2025-05-29
update branch where snapshot can be released from

saperi22:
### Summary of changes

 - Update branch where snapshot can be released from.

 ### Checklist

 - [ ] Added a changelog entry

-


.github/workflows/release_snapshot.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
```
tdchow, line 6: Should we also update this to publish snapshots off of `main` instead of `master`?

jaxdesmarais (approved):

PR #100, opened 2026-03-04
Add PayPal app-switch support to Popup Bridge

srikanthdodda533:

### Overview
This PR adds app-switch support and related analytics to the Popup Bridge Android SDK so that when the PayPal (or Venmo) app is installed, the WebView flow can open the native app and handle the return via deep link. The existing browser-based flow is unchanged. App switch is controlled by the `enablePopupBridgeAppSwitch` parameter (defaults to `false`).

---

#### Library – Core Behavior

| File | Change |
|------|--------|
| **AndroidManifest.xml** | Added `<queries>` for `com.paypal.android.p2pmobile` and `com.venmo` (Android 11+ package visibility). |
| **PopupBridgeAnalytics.kt** | Added 4 new app-switch analytics events: `APP_DETECTED`, `APP_LAUNCHED`, `APP_LAUNCH_FAILED`, `APP_SWITCH_RETURNED`. |
| **AppInstalledChecks.kt** | Added `Context.isPayPalInstalled()` extension (same pattern as `isVenmoInstalled()`). |
| **PopupBridgeJavascriptInterface.kt** | Added `Context` and `enablePopupBridgeAppSwitch` to constructor; exposed `isPayPalInstalled`, `isVenmoInstalled`, and `launchApp(url)` to JS. `isPayPalInstalled` only returns `true` when app switch is enabled. |
| **PopupBridgeWebViewClient.kt** | Removed `setVenmoInstalled()` injection in `onPageFinished`; installed state is now exposed via the JS interface only. |
| **PopupBridgeClient.kt** | See details below. |

**PopupBridgeClient.kt details:**
- **Enable/disable app switch:** New `enablePopupBridgeAppSwitch: Boolean = false` parameter on both the internal and public constructors. Passed through to the JS interface. When enabled, `isPayPalInstalled` returns the actual installed state to JS and `APP_DETECTED` analytics fires at init. When disabled (default), app switch is not available and the existing browser flow is used.
- **App switch:** `launchApp(url)` posts to main thread via `mainHandler`, then `launchAppOnMainThread` builds an `ACTION_VIEW` intent. For PayPal app-switch URLs (`paypal.com/app-switch-checkout*`), the intent is pinned to the PayPal package via `setPackage()`. On `ActivityNotFoundException`, sends `APP_LAUNCH_FAILED` and falls back to `openUrl`.
- **Return handling:** `handleReturnToApp` uses `Uri.isAppSwitchReturnUri()` to distinguish app-switch returns (paths like `onApprove`, `onCancel`, `onError`, `/approve`, `/cancel`, `/error`, or non-empty fragments) from browser-switch returns (query-only, no app-switch path). App-switch returns are handled via `handleAppSwitchReturn`; browser-switch returns fall through to the existing `BrowserSwitchClient` logic.
- **Stale intent cleanup:** `clearPopupBridgeReturnIntentIfPresent()` clears `activity.intent.data` after consuming a popupbridge return, preventing stale intents from being reprocessed on `onResume`. Called after app-switch return, browser-switch completion, and before launching a new app switch.
- **Init:** When app switch is enabled and PayPal is installed, sends `POPUP_BRIDGE_APP_DETECTED`; wires `onLaunchApp`; passes `applicationContext` and `enablePopupBridgeAppSwitch` into the JS interface.
- **Robustness:** Null/blank URL and invalid URI in `launchAppOnMainThread` are rejected and reported via `errorListener`.
- **Companion:** `activeInstance` (`WeakReference`) added for app-switch return handling.

#### Demo App

| File | Change |
|------|--------|
| **PopupActivity.java** | Added `setIntent(newIntent)` in `onNewIntent` so `getIntent()` reflects the deep-link intent. |

#### Tests

| File | Change |
|------|--------|
| **AppInstalledChecksTest.kt** *(new)* | `isPayPalInstalled` / `isVenmoInstalled` – installed and not-installed cases. |
| **PopupBridgeAnalyticsTest.kt** *(new)* | Asserts all 8 analytics constants (4 existing + 4 new). |
| **PopupBridgeClientUnitTest.kt** | App-switch and `launchApp` flows: init analytics, callback wiring, return handling (cancel/approve), success/failure/fallback, null/blank URL, main-thread posting, PayPal package pinning, stale intent cleanup, browser-switch vs app-switch URI disambiguation. |
| **PopupBridgeJavascriptInterfaceUnitTest.kt** | Updated constructor with `Context`; added tests for `isPayPalInstalled`, `isVenmoInstalled`, `launchApp` (valid and null). |
| **PopupBridgeWebViewClientTest.kt** | Removed old Venmo-injection tests; asserts `onPageFinished` does not inject JS. |

---

### Enable/Disable App Switch

The `enablePopupBridgeAppSwitch` parameter (defaults to `false`) controls whether app switch is available:

| When disabled (`false`, default) | When enabled (`true`) |
|----------------------------------|-----------------------|
| `isPayPalInstalled` returns `false` to JS | `isPayPalInstalled` returns the actual installed state |
| `APP_DETECTED` analytics not sent at init | `APP_DETECTED` sent if PayPal app is installed |
| `launchApp(url)` still callable but JS won't trigger it (since `isPayPalInstalled` is `false`) | Full app-switch flow enabled |

This is specific to the popup bridge flow and is separate from the JS SDK's `appSwitchWhenAvailable` which controls non-webview mobile browser app switch.

---

### Merchant Impact
- **New optional parameter:** `enablePopupBridgeAppSwitch` on `PopupBridgeClient` constructor (defaults to `false`). Merchants opt in by passing `true`.
- Merchants still call `handleReturnToApp(intent)` from `onNewIntent` (and optionally `onResume`).
- They must call `setIntent(newIntent)` in `onNewIntent` (existing requirement, unchanged).

### Testing
- [x] `./gradlew :PopupBridge:testDebugUnitTest` – all unit tests pass

### Checklist
- [ ] Added a changelog entry

### Authors
- @srdodda

### DIFF
--- a/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
+++ b/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
@@ -107,7 +107,7 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
 
         with(popupBridgeJavascriptInterface) {
             onOpen = { url -> openUrl(url) }
-            onLaunchApp = { url -> launchApp(url) }
+            onLaunchApp = { url -> this@PopupBridgeClient.launchApp(url) }
             onSendMessage = { messageName, data ->
                 messageListener?.onMessageReceived(messageName, data)
             }
@@ -168,6 +168,11 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
     }
 
     private fun launchAppOnMainThread(url: String?, activity: ComponentActivity) {
+        if (url.isNullOrBlank()) {
+            errorListener?.onError(IllegalArgumentException("Invalid URL for app launch"))
+            return
+        }
+
         activeInstance = WeakReference(this)
         isHandlingReturnToApp = true
 
--- a/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
+++ b/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
@@ -1,8 +1,11 @@
 package com.braintreepayments.api
 
 import android.annotation.SuppressLint
+import android.content.ActivityNotFoundException
 import android.content.Intent
 import android.net.Uri
+import android.os.Handler
+import android.os.Looper
 import android.webkit.WebView
 import androidx.activity.ComponentActivity
 import androidx.core.net.toUri
@@ -15,6 +18,7 @@ import com.braintreepayments.api.internal.AnalyticsParamRepository
 import com.braintreepayments.api.internal.PendingRequestRepository
 import com.braintreepayments.api.internal.PopupBridgeJavascriptInterface
 import com.braintreepayments.api.internal.PopupBridgeJavascriptInterface.Companion.POPUP_BRIDGE_URL_HOST
+import com.braintreepayments.api.internal.isPayPalInstalled
 import java.lang.ref.WeakReference
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
@@ -34,10 +38,14 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
         coroutineScope = activity.lifecycleScope,
     ),
     analyticsParamRepository: AnalyticsParamRepository = AnalyticsParamRepository.instance,
-    popupBridgeJavascriptInterface: PopupBridgeJavascriptInterface = PopupBridgeJavascriptInterface(returnUrlScheme),
+    popupBridgeJavascriptInterface: PopupBridgeJavascriptInterface = PopupBridgeJavascriptInterface(
+        returnUrlScheme = returnUrlScheme,
+        context = activity.applicationContext,
+    ),
 ) {
     private val activityRef = WeakReference(activity)
     private val webViewRef = WeakReference(webView)
+    private val mainHandler = Handler(Looper.getMainLooper())
 
     private var navigationListener: PopupBridgeNavigationListener? = null
     private var messageListener: PopupBridgeMessageListener? = null
@@ -93,8 +101,13 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
         webView.addJavascriptInterface(popupBridgeJavascriptInterface, POPUP_BRIDGE_NAME)
         webView.webViewClient = popupBridgeWebViewClient
 
+        if (activity.applicationContext.isPayPalInstalled()) {
+            analyticsClient.sendEvent(PopupBridgeAnalytics.POPUP_BRIDGE_APP_DETECTED)
+        }
+
         with(popupBridgeJavascriptInterface) {
             onOpen = { url -> openUrl(url) }
+            onLaunchApp = { url -> launchApp(url) }
             onSendMessage = { messageName, data ->
                 messageListener?.onMessageReceived(messageName, data)
             }
@@ -114,6 +127,13 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
     }
 
     fun handleReturnToApp(intent: Intent) {
+        // Try app switch return first (deep link from PayPal/Venmo app)
+        val returnUri = intent.data
+        if (returnUri != null && returnUri.host == POPUP_BRIDGE_URL_HOST) {
+            handleAppSwitchReturn(returnUri)
+            return
+        }
+
         if (isHandlingReturnToApp) return
         isHandlingReturnToApp = true
 
@@ -134,6 +154,53 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
                 is BrowserSwitchFinalResult.NoResult -> runCanceledJavaScript()
             }
             isHandlingReturnToApp = false
+        }
+    }
+
+    /**
+     * Entry point when the web page calls window.popupBridge.launchApp(url).
+     * Posts to the main thread so [startActivity] is never called from a background thread
+     * (JavascriptInterface callbacks can be invoked off the main thread).
+     */
+    private fun launchApp(url: String?) {
+        val activity = activityRef.get() as? ComponentActivity ?: return
+        mainHandler.post { launchAppOnMainThread(url, activity) }
+    }
+
+    private fun launchAppOnMainThread(url: String?, activity: ComponentActivity) {
+        activeInstance = WeakReference(this)
+        isHandlingReturnToApp = true
+
+        val uri = url?.toUri() ?: run {
+            errorListener?.onError(IllegalArgumentException("Invalid URL for app launch"))
+            return
+        }
+
+        val intent = Intent(Intent.ACTION_VIEW, uri).apply {
+            addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+        }
+
+        try {
+            activity.startActivity(intent)
+            analyticsClient.sendEvent(PopupBridgeAnalytics.POPUP_BRIDGE_APP_LAUNCHED)
+        } catch (_: ActivityNotFoundException) {
+            isHandlingReturnToApp = false
+            analyticsClient.sendEvent(PopupBridgeAnalytics.POPUP_BRIDGE_APP_LAUNCH_FAILED)
+            openUrl(url)
+        }
+    }
+
+    private fun handleAppSwitchReturn(returnUri: Uri) {
+        if (!isHandlingReturnToApp) return
+        isHandlingReturnToApp = false
+
+        analyticsClient.sendEvent(PopupBridgeAnalytics.POPUP_BRIDGE_APP_SWITCH_RETURNED)
+
+        val path = returnUri.path.orEmpty()
+        if (path.contains("onCancel")) {
+            runCanceledJavaScript()
+        } else {
+            runNotifyCompleteJavaScript(returnUri)
         }
     }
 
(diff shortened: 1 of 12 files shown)


PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -15,29 +18,38 @@ import com.braintreepayments.api.internal.AnalyticsParamRepository
...
 import com.braintreepayments.api.internal.PopupBridgeJavascriptInterface
 import com.braintreepayments.api.internal.PopupBridgeJavascriptInterface.Companion.POPUP_BRIDGE_URL_HOST
+import com.braintreepayments.api.internal.isPayPalInstalled
 import java.lang.ref.WeakReference
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 import org.json.JSONException
 import org.json.JSONObject
 
+private const val PAYPAL_APP_PACKAGE = "com.paypal.android.p2pmobile"
```
ayer-ribeiro: Should we reuse the constant defined in [AppInstalledChecks.kt](https://github.com/braintree/popup-bridge-android/pull/100/changes#diff-3597dca5b310b9df2239a4a7203c9eefaa18520d99dfc4c0a88d1cc080614293) to keep a single source of truth for the PayPal package name?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -133,14 +166,103 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
...
+     * (JavascriptInterface callbacks can be invoked off the main thread).
+     */
+    private fun launchApp(url: String?) {
+        val activity = activityRef.get() as? ComponentActivity ?: run {
+            return
+        }
+        mainHandler.post { launchAppOnMainThread(url, activity) }
+    }
+
+    private fun launchAppOnMainThread(url: String?, activity: ComponentActivity) {
```
ayer-ribeiro: Right now, `launchAppOnMainThread` just launches the app, while `launchApp` runs the app on the main thread.
This looks inverted. launchApp could be the default entry point, and the other function could explicitly indicate that it must be executed on the main thread.

```suggestion
    private fun launchAppAssumingMainThread(url: String?, activity: ComponentActivity) {
```

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -133,14 +166,103 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
...
+        if (!isHandlingReturnToApp) {
+            return
+        }
+        isHandlingReturnToApp = false
+
+        analyticsClient.sendEvent(PopupBridgeAnalytics.POPUP_BRIDGE_APP_SWITCH_RETURNED)
+        clearPopupBridgeReturnIntentIfPresent("app_switch_return_consumed")
+
+        val path = returnUri.path.orEmpty()
+        if (path.contains("onCancel")) {
```
ayer-ribeiro: `isAppSwitchReturnUri` seems to follow different logic from this function. If the URI contains `/cancel`, `handleAppSwitchReturn` ends up executing the completion flow instead of the cancellation flow.

Maybe we could create a small helper to check whether the URI is a cancel URL and reuse it in `isAppSwitchReturnUri` as well.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -133,14 +166,103 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
...
+
+    private fun launchAppOnMainThread(url: String?, activity: ComponentActivity) {
+        if (url.isNullOrBlank()) {
+            errorListener?.onError(IllegalArgumentException("Invalid URL for app launch"))
+            return
+        }
+
+        clearPopupBridgeReturnIntentIfPresent("launching_new_app_switch")
+        activeInstance = WeakReference(this)
+        isHandlingReturnToApp = true
```
ayer-ribeiro: If I understand correctly, `handleReturnToApp` is called when Popup Bridge needs to handle the return from PayPal/Venmo to the merchant app.
In that case, does it make sense to set i`sHandlingReturnToApp = true` when launching checkout?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -133,14 +166,103 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
...
+            normalizedPath.contains("/approve") ||
+            normalizedPath.contains("/cancel") ||
+            normalizedPath.contains("/error")
+    }
+
+    private fun handleAppSwitchReturn(returnUri: Uri) {
+        if (!isHandlingReturnToApp) {
+            return
+        }
+        isHandlingReturnToApp = false
```
ayer-ribeiro: Shouldn't this be the opposite? When handleAppSwitchReturn is called, I would expect `isHandlingReturnToApp` to be set to true to avoid race condition issues and set isHandlingReturnToApp to `false` when handling is complete. Am I missing anything?

```suggestion
    private fun handleAppSwitchReturn(returnUri: Uri) {
        if (isHandlingReturnToApp) {
            return
        }
        isHandlingReturnToApp = true
```

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -230,6 +368,9 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
         private const val REQUEST_CODE: Int = 1
         private const val POPUP_BRIDGE_NAME: String = "popupBridge"
 
+        @Volatile
+        private var activeInstance: WeakReference<PopupBridgeClient>? = null
+
```
ayer-ribeiro: Is this needed? I didn't find an usage of `activeInstance`.
```suggestion
```

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -133,14 +166,103 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
...
+        analyticsClient.sendEvent(PopupBridgeAnalytics.POPUP_BRIDGE_APP_SWITCH_RETURNED)
+        clearPopupBridgeReturnIntentIfPresent("app_switch_return_consumed")
+
+        val path = returnUri.path.orEmpty()
+        if (path.contains("onCancel")) {
+            runCanceledJavaScript()
+        } else {
+            runNotifyCompleteJavaScript(returnUri)
+        }
```
ayer-ribeiro, line 274: Should `isHandlingReturnToApp` be reset after finishing handling return to app?
```suggestion
        }
        isHandlingReturnToApp = false
```

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
srikanthdodda533: Setting isHandlingReturnToApp = true in launchAppAssumingMainThread feels wrong since handleReturnToApp is what handles returns, not launch. But it's actually needed as a guard so that when the app returns, handleAppSwitchReturn knows a launch is pending. Let me add a comment to clarify, and fix the cancel check inconsistency.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
srikanthdodda533: It's already reset — `isHandlingReturnToApp = false` is set at the top of `handleAppSwitchReturn `(line 267, right after the guard check), before the cancel/complete JS runs. So by the time we reach line 257, the flag is already `false`.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
srikanthdodda533: Thanks — agreed. We added isCancelUri() and use it from both hasAppSwitchPath() (so isAppSwitchReturnUri still recognizes cancel URLs) and handleAppSwitchReturn (so cancel vs complete stays consistent). There’s also a unit test for a path that only matches via /cancel.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
srikanthdodda533: We shouldn’t set this flag to true inside handleAppSwitchReturn. It isn’t a re-entrancy lock; it’s “we launched app switch from launchApp and are waiting for that return.” It’s set true before startActivity and cleared when we handle the matching deep link so we ignore stale popupbridgev1 URLs. Setting true here would flip that model and accept returns we didn’t initiate. We renamed it to expectingAppSwitchReturn to separate it from isHandlingReturnToApp, which is the browser-switch single-flight guard.

noguier:
Hi, I was trying to test these changes on the device with PP installed, and I was not able to app switch, even when i force `enablePopupBridgeAppSwitch` to be true. Will you be able to record a successful path and attach to this PR if app-switch works on your device? Ideally, we would want to test it with the Demo app before merging in.

jaxdesmarais:
@srikanthdodda533 please resolve conflicts

jaxdesmarais:
Moving this to a draft until conflicts are resolved

pedrofsn:
moved to https://github.com/braintree/popup-bridge-android/pull/109

PR #104, opened 2026-04-17
feat: Add static analysis with android lint and detekt

noguier:

### Summary of changes
- add static analysis ruleset 
- add an automatic CI trigger for PR and releases
- fix linter issues triggered by linter ruleset

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Minimal assistance in fixing lint issues discovered in the repo

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry
- [ ] Tested and confirmed payment flows affected by this change are functioning as expected

@noguier


ci
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+command_name="$1"
+
+case $command_name in
+  lint)
+    ./gradlew clean lint
+    ./gradlew detekt --auto-correct
```
saperi22, line 8: `--auto-correct` does this flag auto correct changes?
Typically, in a CI, you only want to check stuff and not modify stuff. Even if it gets modified, the changes aren't being committed, are they?

saperi22 (approved):

ci
noguier: oh good catch , I was following the way lint is set up in the Core BT SDK repo. Let me double check my understanding.

noguier: I added a separate command for GitHub Actions to run without auto-correct, to be safe

saralvasquez (approved):

Issue #95, opened 2025-08-04
v5.0.0 silently overrides WebViewClient


seviraristocrat:
The following code [PopupBridgeClient.kt L91-L96](https://github.com/braintree/popup-bridge-android/blob/main/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt#L91-L96) unconditionally sets a WebViewClient, which causes problems for library users who also need to provide their own WebViewClient for the same WebView (e.g., to handle navigation or logging).

This creates a conflict:
	•	If the user sets their WebViewClient before initializing PopupBridge, the library will override it.
	•	If the user sets their WebViewClient after, it will override PopupBridge’s, and setVenmoInstalled(activity.isVenmoInstalled()) will never be called.

This makes it impossible to use both implementations cleanly.
Please consider exposing a way to pass a custom WebViewClient, or wrapping/delegating to one, so that user callbacks and library logic can coexist.

Code in question PopupBridgeClient.kt L91-L96:
```kotlin
webView.webViewClient = object : WebViewClient() {
    override fun onPageFinished(view: WebView, url: String) {
        super.onPageFinished(view, url)
        setVenmoInstalled(activity.isVenmoInstalled())
    }
}
```

PR #109, opened 2026-05-08
feat: Add Popup Bridge++ app switch for PayPal/Venmo detection and launchApp()

pedrofsn:
### Summary of changes

- Add Popup Bridge++ app switch feature for native PayPal/Venmo app detection
- Add `launchApp()` method to open native PayPal/Venmo apps instead of browser
- Add deep-link return handling for app switch flow
- Add `isPayPalInstalled` and `isVenmoInstalled` JavaScript interface properties
- Enable PayPal native app switch support by default (no opt-in flag required)

> [!NOTE] 
> The `dependency-review` CI step is expected to fail for inner-sourced PRs and is not required for mergeability. Confirmed by Sara Vasquez: this step cannot run on forks due to GitHub permission restrictions.

### Steps to test

> [!TIP]
> **Steps to test:** Click in 'Launch Popupbridge' in the Demo App → PayPal (via Braintree) → Click in PayPal button → initiate a PayPal checkout in **Sandbox mode** → confirm app switch to Venice/Oslo/Venmo native app and return to merchant works end-to-end. 
> 
> For Oslo and Venice, the apps requires to be in sandbox mode (more details [here](https://github.com/braintree/popup-bridge-android/pull/109#issuecomment-4729998829))

#### High-level

```mermaid
sequenceDiagram
    participant User
    participant Demo as Demo App
    participant TestPage as Test Page
    participant Native as Venmo/Oslo/Venice App


    Note over Native: Oslo and Venice must be in Sandbox mode

    User->>Demo: Open Demo App
    Demo->>TestPage: Open the test page
    User->>TestPage: Initiate App Switch
    TestPage->>Native: Trigger App Switch
    Native-->>TestPage: Return via App Switch
    Native->>User: Checkout Completed
```

### AI Usage

**Which AI Agent Was Used?**
- [x] Opencode
- [x] Claude

**How was AI used?**
- Code generation based on existing implementation from srikanthdodda533 fork
- Debugging test failures and resolving merge conflicts
- Writing and fixing tests

- [x] 60 - 100%

### Checklist

- [x] Added a changelog entry
- [x] Tested and confirmed payment flows affected by this change are functioning as expected

### Authors

- [pedrofsn](https://github.com/pedrofsn)
- srikanthdodda533

----
### Inner Source Process

1. Comment `/inner source` on this PR — adds `inner source` and `tech lead review required` labels. Open PR in draft state.
2. PR reviewed and approved by tech lead (no LGTM — comments and feedback required).
3. Comment `/ready` — removes `tech lead review required` label. Move to ready to review.
4. PR comments must be addressed within 24 hours; if unable, move back to draft.

### Inner Source Checklist

- [x] Added all labels to the PR
- [x] Provide steps to test the flows changed (see Summary above)
- [x] [Demo video of the functionality](https://paypal.atlassian.net/wiki/spaces/Checkout/pages/2964575086/Venmo+Popup+Bridge+Testing)
- [x] All upstream dependencies are merged in and this PR can be released at any time
- [x] Unit tests and builds have been run locally and pass/compile as expected


noguier:
Please fix Lint, and instrumentation tests for this PR.

saralvasquez:
Once you've resolved the instrumentation test failure please have your tech lead review this PR. Once it's approved by them you can remove the `tech lead review required` label and that will inform us that it's time for our team's review. Thanks!

pedrofsn:
**Race Condition Fix**

The dismiss test (`testDismissPopupBridge`) was failing consistently on API 35 in CI and locally on arm64, while passing on API 31. This was a classic race condition: the test was asserting immediately after clicking "I don't like any of these colors" without waiting for the full popup bridge return cycle to complete (deep link → onNewIntent → handleReturnToApp → window.popupBridge.onComplete → DOM update). On faster emulators/API levels, the timing worked out; on slower ones it didn't.

**Fix:** Added  before the final assertion, which is consistent with how the color selection tests already handle the same async return flow. This ensures we wait for the bridge to complete before asserting.

PopupBridge/src/main/java/com/braintreepayments/api/internal/PopupBridgeJavascriptInterface.kt
```diff
@@ -1,20 +1,37 @@
...
     private val returnUrlScheme: String,
+    private val context: Context,
+    private val enablePopupBridgeAppSwitch: Boolean = false,
 ) {
 
     var onOpen: ((url: String?) -> Unit)? = null
+    var onLaunchApp: ((url: String?) -> Unit)? = null
     var onSendMessage: ((messageName: String?, data: String?) -> Unit)? = null
 
+/** Exposed to JS as window.popupBridge.isPayPalInstalled. Called by the web page, not by Kotlin. */
```
anibalb2500: nit: Can we fix indentation here

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
```diff
@@ -0,0 +1,127 @@
...
+
+    private fun Uri.isAppSwitchReturnUri(): Boolean {
+        if (host != POPUP_BRIDGE_URL_HOST) return false
+        if (!fragment.isNullOrBlank()) return true
+        return hasAppSwitchPath()
+    }
+
+    private fun Uri.isCancelUri(): Boolean {
+        val normalizedPath = path.orEmpty().lowercase()
+        return normalizedPath.contains("oncancel") || normalizedPath.contains("/cancel")
```
anibalb2500: Any reason why we check `"oncancel"` with no `'/'` while we check with "`/cancel"` with `'/'`?

Do we have an idea of what the expected cancel uri's are going to look like?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -93,9 +111,24 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
         webView.settings.javaScriptEnabled = true
         webView.addJavascriptInterface(popupBridgeJavascriptInterface, POPUP_BRIDGE_NAME)
         webView.webViewClient = popupBridgeWebViewClient
+        if (enablePopupBridgeAppSwitch) {
+            popupBridgeWebViewClient.onVenmoUrl = { url -> appSwitchHandler.launchApp(url) }
+        }
+
+        if (enablePopupBridgeAppSwitch && activity.applicationContext.isPayPalInstalled()) {
```
anibalb2500: nit: Could we do include this conditional in the same `if (enablePopupBridgeAppSwitch)` above?

```
if (enablePopupBridgeAppSwitch) {
    popupBridgeWebViewClient.onVenmoUrl = { url -> appSwitchHandler.launchApp(url) }
    if (activity.applicationContext.isPayPalInstalled()) {
        analyticsClient.sendEvent(PopupBridgeAnalytics.POPUP_BRIDGE_APP_DETECTED)
   }
}
```

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
anibalb2500: From @pedrofsn 
```
The asymmetry is intentional. The two patterns handle two different URI shapes:

"oncancel" matches scheme://popupbridgev1/onCancel, which is what the PayPal/Venmo native app returns (it appends the JS SDK callback name to returnUrlPrefix)
"/cancel" matches scheme://popupbridgev1/some/cancel/flow, where cancel appears as an embedded path segment — the leading slash is a boundary guard to avoid substring false-positives


The same dual-pattern applies to approve/error. That said, we could add "/" to the oncancel/onapprove/onerror checks too ("/oncancel") for consistency — the path always starts with / so it's a safe change that makes the intent clearer.
```

anibalb2500 (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 * Bump `browser-switch` to 3.5.1
 * Upgrade `compileSdkVersion` to API 36
 
+## 5.1.1
+
+* Add Popup Bridge++ app switch: PayPal/Venmo detection, launchApp(), and deep-link return handling
```
jaxdesmarais: looks like there was a merge issue here. Also Popup Bridge++ means nothing to merchants. This is a merchant facing changelog, let's update this to reflect what the functionality updates are and how to use it.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -162,7 +203,9 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
     }
 
     private fun runNotifyCompleteJavaScript(returnUri: Uri) {
-        if (returnUri.host != POPUP_BRIDGE_URL_HOST) return
+        if (returnUri.host != POPUP_BRIDGE_URL_HOST) {
+            return
+        }
```
jaxdesmarais: Why this change? Looks like logic is the exact same.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -65,20 +77,26 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
      * @param webView The [WebView] to enable for PopupBridge.
      * @param returnUrlScheme The return url scheme to use for deep linking back into the application.
      * @param popupBridgeWebViewClient The [PopupBridgeWebViewClient] to use for handling web view events.
+     * @param enablePopupBridgeAppSwitch When true, allows the SDK to launch the native PayPal app
```
jaxdesmarais: Lets align param names across iOS and Android and make this specific to PayPal app switch.

Demo/src/main/java/com/braintreepayments/popupbridge/demo/MainActivity.java
```diff
@@ -14,6 +13,7 @@ public class MainActivity extends AppCompatActivity {
     private static final String PAYPAL_CHECKOUTJS_POPUP_BRIDGE_URL = "https://braintree.github.io/popup-bridge-example/paypal-checkout.html";
     private static final String VENMO_POPUP_BRIDGE_URL = "https://braintree.github.io/popup-bridge-example/venmo";
     private static final String LPM_POPUP_BRIDGE_URL = "https://braintree.github.io/popup-bridge-example/local-payment-methods";
+    public static final String CUSTOM_URL = "https://paypal.com";
```
saralvasquez: Why is this named `CUSTOM_URL`? That naming doesn't feel very descriptive

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeAnalytics.kt
```diff
@@ -5,4 +5,8 @@ internal object PopupBridgeAnalytics {
     const val POPUP_BRIDGE_SUCCEEDED = "popup-bridge:succeeded"
     const val POPUP_BRIDGE_FAILED = "popup-bridge:failed"
     const val POPUP_BRIDGE_CANCELED = "popup-bridge:canceled"
+    const val POPUP_BRIDGE_APP_DETECTED = "popup-bridge:app-switch:app-detected"
+    const val POPUP_BRIDGE_APP_LAUNCHED = "popup-bridge:app-switch:app-launched"
+    const val POPUP_BRIDGE_APP_LAUNCH_FAILED = "popup-bridge:app-switch:app-launch-failed"
+    const val POPUP_BRIDGE_APP_SWITCH_RETURNED = "popup-bridge:app-switch:returned"
```
jaxdesmarais: Let's align analytics and naming across platforms

CHANGELOG.md
noguier: Additionally to Jax's suggestion, please add a `CHANGELOG` entry under `##unreleased`

noguier:
When testing the flows, even with `enablePopupBridgeAppSwitch` set to true, I am unable to app switch to PayPal. i can appswitch to Venmo

saralvasquez:
> When testing the flows, even with `enablePopupBridgeAppSwitch` set to true, I am unable to app switch to PayPal. i can appswitch to Venmo

I'm experiencing the same behavior

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Bump `browser-switch` to 3.5.1
 * Upgrade `compileSdkVersion` to API 36
+* Add opt-in PayPal native app switch support. Pass `enablePayPalAppSwitch = true` to `PopupBridgeClient` to launch the PayPal or Venmo app directly instead of opening a browser tab when available. The result is returned via the existing `handleReturnToApp(intent)` call with no additional integration changes required. Defaults to `false` for full backward compatibility.
```
saralvasquez: This is an incredibly long CHANGELOG entry. This doesn't have to be a full detailed explanation of that changes just a note so that users knows what changed in this version. You can probably just keep the first sentence

Demo/src/main/res/layout/activity_main.xml
```diff
@@ -45,5 +45,12 @@
         android:onClick="onLPMPopupBridgeClick"
         android:text="Launch Local Payment Methods" />
 
+    <Button
+        android:id="@+id/button_custom_url"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:contentDescription="Launch custom URL"
+        android:onClick="onCustomURLClick"
+        android:text="Launch custom URL" />
```
saralvasquez: What is the point of adding this? I'm not sure what it has to do with the features being added in this branch

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
```diff
@@ -0,0 +1,127 @@
...
+        val normalizedPath = path.orEmpty().lowercase()
+        return normalizedPath.contains("onapprove") ||
+            normalizedPath.contains("onerror") ||
+            normalizedPath.contains("/approve") ||
+            normalizedPath.contains("/error") ||
+            isCancelUri()
+    }
+}
+
+internal fun Uri.isVenmoAppSwitchUri(): Boolean {
```
saralvasquez, line 127: Just out of curiosity, why is Venmo handled so differently than PayPal? It's clear that this is used in other areas of the code and the PayPal version of this logic is not, but why is that necessary?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -65,20 +77,26 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
...
      * @throws IllegalArgumentException If the activity is not valid or the fragment cannot be added.
      */
     @JvmOverloads
     constructor(
         activity: ComponentActivity,
         webView: WebView,
         returnUrlScheme: String,
-        popupBridgeWebViewClient: PopupBridgeWebViewClient = PopupBridgeWebViewClient()
+        popupBridgeWebViewClient: PopupBridgeWebViewClient = PopupBridgeWebViewClient(),
+        enablePayPalAppSwitch: Boolean = false,
```
saralvasquez, line 98: Why is there no matching venmo component to the PayPal specific pieces in this functionality?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -93,9 +111,23 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
         webView.settings.javaScriptEnabled = true
         webView.addJavascriptInterface(popupBridgeJavascriptInterface, POPUP_BRIDGE_NAME)
         webView.webViewClient = popupBridgeWebViewClient
+        if (enablePayPalAppSwitch) {
+            popupBridgeWebViewClient.onVenmoUrl = { url -> appSwitchHandler.launchApp(url) }
+            popupBridgeJavascriptInterface.onLaunchApp = { url -> appSwitchHandler.launchApp(url) }
+            if (activity.applicationContext.isPayPalInstalled()) {
+                analyticsClient.sendEvent(PopupBridgeAnalytics.POPUP_BRIDGE_APP_DETECTED)
```
saralvasquez: Again there is a lot of functionality here that is PayPal specific without there being a venmo counterpart, why is that?

saralvasquez (commented):
This is a great first pass! I have a few comments and questions. Additionally I ran a quick search with Claude and it seems like there might be some logic paths that aren't covered by unit tests. Could you just take a quick pass at these flows and double check they're covered. Thanks!

`  1. PopupBridgeWebViewClient Venmo URL interception — The new shouldOverrideUrlLoading block that intercepts Venmo URLs and invokes onVenmoUrl has no test. No test sets onVenmoUrl on the client, passes a
  https://account.venmo.com/braintree/checkout request, and verifies the callback fires and true is returned.`
`2. onOpen routing Venmo URLs to app switch — When enablePayPalAppSwitch=true, the onOpen lambda in PopupBridgeClient checks uri.isVenmoAppSwitchUri() and routes to appSwitchHandler.launchApp() instead  
  of openUrl(). This branch is untested.`
`3. isPayPalInstalled JS property when app IS installed — The existing test only covers enablePayPalAppSwitch=false (asserts false). There is no test for enablePayPalAppSwitch=true with PayPal actually  
  installed (asserting true).`
`4. Fragment-based app-switch return URI — isAppSwitchReturnUri returns true when host == POPUP_BRIDGE_URL_HOST && !fragment.isNullOrBlank(). No test exercises a return URI that uses a fragment instead  
  of a recognized path.`
`5. Sandbox PayPal URL — isPayPalAppSwitchUri explicitly handles sandbox.paypal.com/app-switch-checkout, but there's no test for it — only paypal.com is exercised indirectly.`
`6. Error path return URIs — hasAppSwitchPath matches /onerror and /error paths and routes them to onComplete. No test verifies this; only /onApprove and cancel paths are covered.`

saralvasquez:
Op also it looks like there's some lint issues here. Could you take a look at that?

CHANGELOG.md
pedrofsn: ok

Demo/src/main/java/com/braintreepayments/popupbridge/demo/MainActivity.java
pedrofsn: I removed it.

Demo/src/main/res/layout/activity_main.xml
pedrofsn: I removed it

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
pedrofsn: The difference is in how each SDK initiates the flow. For Venmo, the PPCP JS SDK triggers a WebView navigation to `account.venmo.com/braintree/checkout` — we intercept that URL in `shouldOverrideUrlLoading` before it renders. The `isVenmoAppSwitchUri` helper lives here (top-level) because `PopupBridgeWebViewClient` also needs it for the interception. The host is rewritten from `account.venmo.com` to `venmo.com` because the Venmo app has no intent filter for the `account` subdomain, only for the root domain.

pedrofsn: For PayPal, the JS SDK calls `window.popupBridge.launchApp(url)` directly with a `paypal.com/app-switch-checkout` URL — no interception or URL rewriting needed. The asymmetry reflects how each SDK surfaces its app switch, not an arbitrary design choice on our side.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
pedrofsn: Venmo is handled here too — when `enablePayPalAppSwitch` is `true`, `popupBridgeWebViewClient.onVenmoUrl` is wired to `appSwitchHandler.launchApp(url)` (a few lines below), so Venmo navigations are intercepted at the WebView level and launched natively.

pedrofsn: The PayPal-specific additions (`isPayPalInstalled` exposed to JS, `onLaunchApp` callback) exist because PayPal's flow is JS-initiated: the SDK reads `window.popupBridge.isPayPalInstalled` before deciding to call `launchApp`. Venmo's flow skips that check entirely — it goes straight to navigating the WebView, which we intercept. So there's no Venmo equivalent of `isVenmoInstalled` in JS because detection isn't part of how Venmo initiates its flow.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
pedrofsn: Same reasoning as above. The `isPayPalInstalled` analytics event here fires because PayPal's JS SDK checks `window.popupBridge.isPayPalInstalled` at page load to decide the checkout path. There's no equivalent Venmo probe because Venmo's SDK doesn't ask — it just navigates, and we catch it in `shouldOverrideUrlLoading`. Different initiation mechanisms led to this asymmetry, not a gap in the implementation.

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
```diff
@@ -0,0 +1,127 @@
...
+        val uri = url.toUri()
+        val targetUri = if (uri.isVenmoAppSwitchUri()) uri.rewriteToVenmoHost() else uri
+        val intent = Intent(Intent.ACTION_VIEW, targetUri).apply {
+            addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+            if (targetUri.isPayPalAppSwitchUri()) {
+                setPackage(PAYPAL_APP_PACKAGE)
+            } else if (uri.isVenmoAppSwitchUri()) {
+                setPackage(VENMO_APP_PACKAGE)
+            }
+        }
```
jaxdesmarais, line 88: When the URL is neither a PayPal nor a Venmo app-switch URI, this still fires Intent(ACTION_VIEW, uri) with FLAG_ACTIVITY_NEW_TASK and no package set, an implicit intent for an arbitrary URI that originates from web content (reachable via `window.popupBridge.launchApp(anyUrl)` once the flag is on). This is the Android version of the validation gap Copilot flagged on iOS #95. We should validate the URI is an expected PayPal/Venmo app-switch URL (https + known host/path) before launching, and otherwise route to `onError/onOpenUrl` rather than launching a bare implicit intent. Note the happy-path test (...invoke("https://www.paypal.com/checkout")) actually exercises this no-package branch, so it's not hypothetical.

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    private fun launchAppOnMainThread(url: String?, activity: ComponentActivity) {
+        if (url.isNullOrBlank()) {
+            onError(IllegalArgumentException("Invalid URL for app launch"))
+            return
+        }
+
+        clearReturnIntentIfPresent()
+        expectingAppSwitchReturn = true
```
jaxdesmarais, line 79: `expectingAppSwitchReturn` is set true before the switch, but there's no path that resets it / signals cancel if the user returns to the merchant app without a deep link (back button, recents, or backgrounding the PayPal app). Browser-switch handles this via `BrowserSwitchFinalResult.NoResult -> runCanceledJavaScript()`; the app-switch path has no equivalent, so the JS flow hangs. This is the same "missing return hangs the flow" failure mode we worried about on iOS, just via a different mechanism. Can we add an app-switch cancel/no-result path (e.g. on `onResume` with no matching deep link while `expectingAppSwitchReturn` is `true`)?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -134,6 +174,7 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
                 )
                 is BrowserSwitchFinalResult.NoResult -> runCanceledJavaScript()
             }
+            appSwitchHandler.clearReturnIntentIfPresent()
```
jaxdesmarais: `appSwitchHandler.clearReturnIntentIfPresent()` runs in the browser-switch coroutine branch unconditionally i.e. even when `enablePayPalAppSwitch` is `false`. It mutates `activity.intent` (nulls data) whenever the host is popupbridgev1, which is true for normal browser-switch returns. That's an un-gated behavior change on the default/disabled path. Please move it inside the `enablePayPalAppSwitch` guard so the feature-off flow is byte-for-byte unchanged.

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
```diff
@@ -0,0 +1,127 @@
...
+            path.orEmpty().startsWith("/app-switch-checkout")
+    }
+
+    private fun Uri.isAppSwitchReturnUri(): Boolean {
+        if (host != POPUP_BRIDGE_URL_HOST) return false
+        if (!fragment.isNullOrBlank()) return true
+        return hasAppSwitchPath()
+    }
+
+    private fun Uri.isCancelUri(): Boolean {
```
jaxdesmarais, line 115: The cancel/approve/error classification uses substring matching (contains("oncancel") || contains("/cancel"), and the same for approve/error in `hasAppSwitchPath`). Legitimate path segments like /cancellation or /error-details would be misrouted, and it's hard to reason about (also the oncancel vs /cancel asymmetry @anibalb2500 already asked about). Can we switch to exact path-segment matching against the known return paths instead of contains?

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
```diff
@@ -0,0 +1,127 @@
...
+        val isPayPal = normalizedHost == "paypal.com"
+        val isSandbox = normalizedHost == "sandbox.paypal.com"
+        return scheme.equals("https", ignoreCase = true) &&
+            (isPayPal || isSandbox) &&
+            path.orEmpty().startsWith("/app-switch-checkout")
+    }
+
+    private fun Uri.isAppSwitchReturnUri(): Boolean {
+        if (host != POPUP_BRIDGE_URL_HOST) return false
+        if (!fragment.isNullOrBlank()) return true
```
jaxdesmarais, line 111: `isAppSwitchReturnUri` treats any non-blank fragment as an app-switch return. That's very broad it's only saved by the `expectingAppSwitchReturn` gate upstream. What return shapes actually carry a fragment here? If it's a specific known case we should match it explicitly rather than "any fragment wins."

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -93,9 +111,23 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
         webView.settings.javaScriptEnabled = true
         webView.addJavascriptInterface(popupBridgeJavascriptInterface, POPUP_BRIDGE_NAME)
         webView.webViewClient = popupBridgeWebViewClient
+        if (enablePayPalAppSwitch) {
```
jaxdesmarais, line 120: Naming: `enablePayPalAppSwitch` gates both PayPal and Venmo here (the `onVenmoUrl` wiring on the next line is inside this block). That's the same concern I raised on iOS, the flag name under-describes its scope. Let's either rename it to reflect that it enables app switch for both, or split the two. (Also worth aligning the final name 1:1 with whatever lands on iOS.)

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -93,9 +111,23 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
...
+            popupBridgeWebViewClient.onVenmoUrl = { url -> appSwitchHandler.launchApp(url) }
+            popupBridgeJavascriptInterface.onLaunchApp = { url -> appSwitchHandler.launchApp(url) }
+            if (activity.applicationContext.isPayPalInstalled()) {
+                analyticsClient.sendEvent(PopupBridgeAnalytics.POPUP_BRIDGE_APP_DETECTED)
+            }
+        }
 
         with(popupBridgeJavascriptInterface) {
-            onOpen = { url -> openUrl(url) }
+            onOpen = { url ->
```
jaxdesmarais, line 126: Venmo app-switch URLs are now intercepted in two places here in the onOpen JS-bridge wrapper, and in `PopupBridgeWebViewClient.shouldOverrideUrlLoading`. Is both paths genuinely needed, or can Venmo interception live in one place? This is the "app-switch logic sprinkled through the coordinator" smell @buzzamus flagged on iOS.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeAnalytics.kt
```diff
@@ -5,4 +5,8 @@ internal object PopupBridgeAnalytics {
     const val POPUP_BRIDGE_SUCCEEDED = "popup-bridge:succeeded"
     const val POPUP_BRIDGE_FAILED = "popup-bridge:failed"
     const val POPUP_BRIDGE_CANCELED = "popup-bridge:canceled"
+    const val POPUP_BRIDGE_APP_DETECTED = "popup-bridge:app-switch:available"
+    const val POPUP_BRIDGE_APP_LAUNCHED = "popup-bridge:app-switch:started"
+    const val POPUP_BRIDGE_APP_LAUNCH_FAILED = "popup-bridge:app-switch:failed"
+    const val POPUP_BRIDGE_APP_SWITCH_RETURNED = "popup-bridge:app-switch:succeeded"
```
jaxdesmarais: Same note I left on iOS: app detection (app-switch:available) should be a tag on the existing events, not its own event event names are for conversion, and the other author removed the standalone detection event on iOS for this reason. Let's keep the analytics shape consistent across platforms.

jaxdesmarais (changes requested):
Thanks for this, the extraction into `AppSwitchHandler` is the right instinct, and I want to call out that Android avoids the biggest issue we hit on iOS (#95): the return comes back through the existing, explicit `handleReturnToApp(intent)` entry point rather than a global NotificationCenter post, so there's no public transport constant and no merchant-side wiring. 👍

That said, a few things carry over from the iOS review and a few are Android-specific. My top concern is the functional report below let's get that sorted before deeper review

Blocker: Two reviewers report PayPal app switch doesn't actually work with the flag enabled (Venmo does). Since native PayPal switch is the headline of this PR, we need a repro/fix before sign-off. Likely suspects: the JS SDK never seeing window.popupBridge.isPayPalInstalled === true, the `com.paypal.android.p2pmobile` package not matching the test build, or the PayPal app lacking an intent filter for the /app-switch-checkout path the intent pins to. Can you confirm which, with the analytics (app-switch:available / :started / :failed) from a failing run?

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
pedrofsn: Good catch. Will add explicit URI validation before `startActivity`: only `https://` URIs matching `paypal.com/sandbox.paypal.com` (path `/app-switch-checkout`) or `venmo.com/account.venmo.com` (path `/braintree/checkout`) are allowed. Anything else routes to `onError`. Will fix the test to also cover the rejection path.

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
pedrofsn: Agreed. I will add a cancel path in `onResume`: when `expectingAppSwitchReturn` is `true` and `shouldHandleReturn(intent.data)` is `false`, we'll call `onCanceled()` and reset the flag. This mirrors `BrowserSwitchFinalResult.NoResult` and ensures the JS flow is never left hanging.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
pedrofsn: Fixed locally. `clearReturnIntentIfPresent()` is now wrapped inside `if (enablePayPalAppSwitch)` (lines 183-185 of PopupBridgeClient.kt). (Fix applied, pending commit/push.) Gap noted: no unit test verifies `clearReturnIntentIfPresent()` is NOT called when feature is off — the feature-off test in PopupBridgeClientUnitTest.kt does not mock or assert on this method call

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
pedrofsn: fragment is the fallback for PayPal return URIs that carry no path (e.g. `popupbridgev1://popupbridgev1#someFragment`, confirmed in `AppSwitchHandlerTest`). 

Logic: host == popupbridgev1 AND (fragment present OR hasAppSwitchPath()). 

The broad fragment check is safe because `expectingAppSwitchReturn` upstream gate only allows this branch after we explicitly launched an app switch. The check is intentional, not a loose guard.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
pedrofsn: Confirmed from source: iOS Braintree SDK uses `enablePayPalAppSwitch` in `BTPayPalVaultRequest.swift` and `BTPayPalCheckoutRequest.swift`. 

Android was previously `enablePopupBridgeAppSwitch` and was renamed in commit `cc404e7` specifically to match iOS.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
pedrofsn: We verified this against the Venmo Android SDK (`checkoutJSIntegration.js`). The PPCP JS SDK calls `window.popupBridge.open(url)`, that is the real trigger. The two paths are not redundant but complementary.

**Path B:** `onOpen` (`PopupBridgeClient.kt:123`): fires first when PPCP JS calls `window.popupBridge.open(venmoUrl)`. popup-bridge intercepts at the JS bridge layer, routes to `appSwitchHandler.launchApp()`, and the WebView navigation never happens. This is the primary path.

**Path A:** `shouldOverrideUrlLoading` (`PopupBridgeWebViewClient.kt:40`): fires if the WebView navigates `window.location directly`, i.e. a code path that bypasses `popupBridge.open()` (different SDK version, or direct navigation before popup bridge initializes). Defensive fallback.

The two paths are mutually exclusive in practice: if Path B fires, `window.location` never changes so Path A never triggers. If the JS navigates directly (Path B missed), Path A catches it. The duplication is intentional defense-in-depth, not an oversight

pedrofsn: Confirmed in `checkoutJSIntegration.js` line 14 does `window.location` = `openUrl`, the JS SDK itself performs a `window.location` assignment. Path A (shouldOverrideUrlLoading) is the direct safety net for that exact assignment if Path B misses. Both paths are necessary

pedrofsn:
> Op also it looks like there's some lint issues here. Could you take a look at that?

done

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeAnalytics.kt
pedrofsn: Solved

pedrofsn:
> Two reviewers report PayPal app switch doesn't actually work with the flag enabled (Venmo does)

It's working with Venmo and also with Oslo. The video below was recorded in a sandbox environment.

https://github.com/user-attachments/assets/0dab18da-b510-4dbc-a56f-efb1eb0e65e5

pedrofsn:
And now, the Venice reproduction in sandbox.

https://github.com/user-attachments/assets/bf230539-07e9-4fd5-b2a5-92d17da234c7

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
pedrofsn: Fixed.

pedrofsn:
could you review it again, please?
@saralvasquez @noguier @jaxdesmarais

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
pedrofsn: Switched to exact segment matching via `split("/")` + `==`. Paths like `/cancellation` or `/error-details` won't be misrouted.

noguier:
Hi, please the video below to see the flow, that I am seeing when testing on the physical device, it behaves the same for both : `enablePayPalAppSwitch = true` , and `enablePayPalAppSwitch = false`. 

https://github.com/user-attachments/assets/5274231b-e6c9-440a-bda8-383de06ccfcb

Is this result expected? 

Additionally, can you please update the PR description with all the changes that were already committed? For example `enablePayPalAppSwitch` used to be `enablePopupBridgeAppSwitch` for both PP and Venmo, the conversation/PR is long enough that it is hard to follow if not listed in the PR description.

pedrofsn:
> Hi, please the video below to see the flow, that I am seeing when testing on the physical device, it behaves the same for both : `enablePayPalAppSwitch = true` , and `enablePayPalAppSwitch = false`.
> 
>  Screen.Recording.2026-06-16.at.4.33.19.PM.mov 
> Is this result expected?
> 
> Additionally, can you please update the PR description with all the changes that were already committed? For example `enablePayPalAppSwitch` used to be `enablePopupBridgeAppSwitch` for both PP and Venmo, the conversation/PR is long enough that it is hard to follow if not listed in the PR description.

Hi @noguier, thanks for the video - that behavior is expected in the context of the Demo app.

The web page loaded by the Demo app is hardcoded to a Sandbox checkout URL, which requires the PayPal app (Venice or Oslo) to be running in Sandbox mode. Without that, both `enablePayPalAppSwitch = true` and `false` will appear identical, as shown in your video.

Unfortunately, the page loaded by the Demo app doesn't have an environment toggle that could be a future improvement to make testing easier.

Regarding the PR description - good call. I'll update it to reflect the current state:
- `enablePayPalAppSwitch` (renamed from `enablePopupBridgeAppSwitch`)
- Scope: gates both PayPal and Venmo app-switch paths
- To test it, your PayPal app must be running in Sandbox mode (Venice or Oslo)

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -93,9 +110,20 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
...
+        if (enablePayPalAppSwitch) {
+            popupBridgeWebViewClient.onVenmoUrl = { url -> appSwitchHandler.launchApp(url) }
+            popupBridgeJavascriptInterface.onLaunchApp = { url -> appSwitchHandler.launchApp(url) }
+        }
 
         with(popupBridgeJavascriptInterface) {
-            onOpen = { url -> openUrl(url) }
+            onOpen = { url ->
+                val uri = url?.toUri()
+                if (enablePayPalAppSwitch && uri != null && uri.isVenmoAppSwitchUri()) {
```
noguier, line 128: can you give a little more context here why we only need to check `uri.isVenmoAppSwitchUri()` here and not PayPal are they the same thing?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeAnalytics.kt
```diff
@@ -5,4 +5,7 @@ internal object PopupBridgeAnalytics {
     const val POPUP_BRIDGE_SUCCEEDED = "popup-bridge:succeeded"
     const val POPUP_BRIDGE_FAILED = "popup-bridge:failed"
     const val POPUP_BRIDGE_CANCELED = "popup-bridge:canceled"
+const val POPUP_BRIDGE_APP_LAUNCHED = "popup-bridge:app-switch:started"
```
noguier: can we match the name of the const val to their value, `const val POPUP_BRIDGE_APP_SWITCH_STARTED = "popup-bridge:app-switch:started"`, etc

Demo/src/main/java/com/braintreepayments/popupbridge/demo/PopupActivity.java
```diff
@@ -25,15 +24,23 @@ protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_popup);
         webView = findViewById(R.id.web_view);
+        String url = getIntent().getStringExtra("url");
+        boolean enableAppSwitch = getIntent().getBooleanExtra("enableAppSwitch", false);
+
+        if (enableAppSwitch) {
+            webView.getSettings().setDomStorageEnabled(true);
+            webView.getSettings().setJavaScriptEnabled(true);
+            CookieManager.getInstance().setAcceptThirdPartyCookies(webView, true);
```
noguier, line 33: why do we need to `acceptThirdPartyCookies` here?

Demo/src/main/java/com/braintreepayments/popupbridge/demo/PopupActivity.java
```diff
@@ -45,14 +52,15 @@ protected void onResume() {
...
     }
 
     public void showDialog(String message) {
         new AlertDialog.Builder(this)
-            .setMessage(message)
-            .setPositiveButton(android.R.string.ok, (dialog, which) -> dialog.dismiss())
-            .show();
+                .setMessage(message)
+                .setPositiveButton(android.R.string.ok, (dialog, which) -> dialog.dismiss())
+                .show();
```
noguier: why do we need the extra indentation here?

pedrofsn: I will take a look

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
```diff
@@ -0,0 +1,134 @@
...
+
+        val uri = url.toUri()
+        if (!uri.isPayPalAppSwitchUri() && !uri.isVenmoAppSwitchUri()) {
+            onError(IllegalArgumentException("URL is not a valid PayPal or Venmo app-switch URI: $url"))
+            return
+        }
+
+        clearReturnIntentIfPresent()
+        expectingAppSwitchReturn = true
+        val targetUri = if (uri.isVenmoAppSwitchUri()) uri.rewriteToVenmoHost() else uri
```
noguier, line 80: I am curious about this logic here, would you mind explaining it to me, including why we need the two extra helpers for venmo?

Demo/src/main/java/com/braintreepayments/popupbridge/demo/PopupActivity.java
pedrofsn: The **Demo app** requires it in order for the PayPal button (and page) to work properly.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -45,17 +52,21 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
     private var errorListener: PopupBridgeErrorListener? = null
 
     /**
-     * Ensures that [handleReturnToApp] is only called once, even if both `onResume` and `onNewIntent` invoke it.
```
noguier, line 48: The kdoc that was removed contained a useful description, can you add that back?

Demo/src/main/java/com/braintreepayments/popupbridge/demo/PopupActivity.java
pedrofsn: The PayPal JS SDK renders checkout content from `paypal.com` inside a cross-origin iframe embedded in the merchant page. From the Webview's perspective, cookies set by that iframe are thirdy-party cookies. And Android block them by default. Without this call, the SDK fails to establish a session and and the button never initialize.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeAnalytics.kt
pedrofsn: Done

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
pedrofsn: PayPal app switch is always initiated via `window.popupBridge.launchApp(url)`. That maps to the `onLaunchApp` callback (line 115), which routes directly to `appswitchHandler.launchApp().` PayPal never calls `popupBridge.open()` for app switch.

pedrofsn: Venmo, on the other hand, calls `window.popupBridge.open(url)` and the implementation in `checkoutJSIntegration.js` immediately does `window.location modifiedUrl`, which triggers `shouldOverrideUrlLoading` in the WebView client. That is the path intercepted at line 121. The two JS APIs handle distinct call paths and are not interchangeable.

Demo/src/main/java/com/braintreepayments/popupbridge/demo/PopupActivity.java
noguier: and, I assume, there is no other way to initialize a button?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeWebViewClient.kt
```diff
@@ -13,17 +13,20 @@ import android.webkit.WebResourceResponse
...
     private val delegate: WebViewClient? = null
 ) : WebViewClient() {
 
+    internal var onVenmoUrl: ((String) -> Unit)? = null
+
     override fun onPageFinished(view: WebView?, url: String?) {
         super.onPageFinished(view, url)
-        setVenmoInstalled(view, view?.context?.isVenmoInstalled() == true)
         delegate?.onPageFinished(view, url)
+        setVenmoInstalled(view, view?.context?.isVenmoInstalled() == true)
```
noguier: is there a reason the  `setVenmoInstalled(view, view?.context?.isVenmoInstalled() == true)` was moved 2 lines down?

PopupBridge/src/test/java/com/braintreepayments/api/PopupBridgeWebViewClientTest.kt
```diff
@@ -45,34 +47,34 @@ class PopupBridgeWebViewClientTest {
     @Test
     fun `on init, when venmo installed, isVenmoInstalled is set to true on the popupBridgeJavascriptInterface`() =
         runTest {
-        every { webView.context.isVenmoInstalled() } returns true
+            every { webView.context.isVenmoInstalled() } returns true
+            val scriptSlot = slot<String>()
 
-        sut.onPageFinished(webView, "https://example.com")
+            sut.onPageFinished(webView, "https://example.com")
```
noguier: nit: i believe the indentation is unnecessary here

PopupBridge/src/test/java/com/braintreepayments/api/PopupBridgeWebViewClientTest.kt
```diff
@@ -45,34 +47,34 @@ class PopupBridgeWebViewClientTest {
...
 
     @Test
     fun `on init, when venmo is not installed, isVenmoInstalled is false on javascriptInterface`() =
         runTest {
-        every { webView.context.isVenmoInstalled() } returns false
+            every { webView.context.isVenmoInstalled() } returns false
+            val scriptSlot = slot<String>()
 
-        sut.onPageFinished(webView, "https://example.com")
+            sut.onPageFinished(webView, "https://example.com")
```
noguier: nit: indentation is unnecessary here

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Bump `browser-switch` to 3.5.1
 * Upgrade `compileSdkVersion` to API 36
```
noguier, line 10: The new version of this library was [released](https://github.com/braintree/popup-bridge-android/releases/tag/5.2.0) 2 days ago, please rebase your branch and the `CHANGELOG` will be updated. There will be an entry for version 5.2.0, your entry, however, should still go under `unreleased`

Demo/src/main/java/com/braintreepayments/popupbridge/demo/MainActivity.java
```diff
@@ -24,6 +23,7 @@ protected void onCreate(Bundle savedInstanceState) {
     private void switchToWebView(String url) {
         Intent intent = new Intent(this, PopupActivity.class);
         intent.putExtra("url", url);
+        intent.putExtra("enableAppSwitch", true);
```
saralvasquez: Will this mean that the demo will always attempt app switch? I'm wondering if it would be helpful to have a configurable setting the demo app to specify if we wanna test the flow with app switch or not. @jaxdesmarais @noguier what do you guys think?

PopupBridge/src/main/java/com/braintreepayments/api/internal/PopupBridgeJavascriptInterface.kt
```diff
@@ -1,34 +1,57 @@
...
- * first argument.
+ * first argument. The web page can then use [window.popupBridge] to access:
+ * - [isPayPalInstalled] / [isVenmoInstalled] — whether the native apps are installed
+ * - [open], [launchApp] — open a URL in browser or native app
+ * - [sendMessage] — send messages to the host
  */
 internal class PopupBridgeJavascriptInterface(
     private val returnUrlScheme: String,
+    private val context: Context,
+    private val enablePayPalAppSwitch: Boolean = false,
```
saralvasquez, line 18: Why is there no enableVenmoAppSwitch?

PopupBridge/src/main/java/com/braintreepayments/api/internal/PopupBridgeJavascriptInterface.kt
```diff
@@ -1,34 +1,57 @@
...
 ) {
 
     var onOpen: ((url: String?) -> Unit)? = null
+    var onLaunchApp: ((url: String?) -> Unit)? = null
     var onSendMessage: ((messageName: String?, data: String?) -> Unit)? = null
 
+    /** Exposed to JS as window.popupBridge.isPayPalInstalled. Called by the web page, not by Kotlin. */
+    @get:JavascriptInterface
+    val isPayPalInstalled: Boolean
+        get() = enablePayPalAppSwitch && context.isPayPalInstalled()
```
saralvasquez, line 28: why is the value of isPayPalInstalled determined by whether it's installed in context and also if app switch is enabled? I don't see how the app switch part is related to whether it's installed or not

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeWebViewClient.kt
saralvasquez: Isn't this already set above in the `PopupBridgeJavascriptInterface`?

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
```diff
@@ -0,0 +1,134 @@
...
+        if (!expectingAppSwitchReturn) return
+        expectingAppSwitchReturn = false
+        onCanceled()
+    }
+
+    fun handleReturn(returnUri: Uri) {
+        if (!expectingAppSwitchReturn) return
+        expectingAppSwitchReturn = false
+
+        analyticsClient.sendEvent(PopupBridgeAnalytics.POPUP_BRIDGE_APP_SWITCH_SUCCEEDED)
```
saralvasquez: Do we want to send a succeeded event if there's a chance to flow was canceled? Would we prefer a different analytic event? This feels more like an analytics preference. @jaxdesmarais thoughts?

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
saralvasquez: I have the same concerns about this name as Jax. I think alignment is important but priority number one feels like a descriptive and easy to understand name

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -115,7 +143,18 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
...
 
     fun handleReturnToApp(intent: Intent) {
-        if (isHandlingReturnToApp) return
+        val returnUri = intent.data
+        if (enablePayPalAppSwitch) {
+            if (appSwitchHandler.shouldHandleReturn(returnUri)) {
+                appSwitchHandler.handleReturn(returnUri!!)
+                return
+            }
+            appSwitchHandler.handleNoResult()
```
saralvasquez, line 159: If I'm reading this correctly this would get hit if there was a app switch on cancel right? Should we return here instead of continuing below?

saralvasquez (changes requested):
I have a couple new questions/concerns and also mirror some of Jax's comments as well

Demo/src/main/java/com/braintreepayments/popupbridge/demo/MainActivity.java
pedrofsn: Renamed

pedrofsn: Currently, the demo app does not have an `enablePayPalAppSwitch` toggle. But we can implement it in another PR.

PopupBridge/src/main/java/com/braintreepayments/api/internal/PopupBridgeJavascriptInterface.kt
pedrofsn: The `enablePayPalAppSwitch` flag was introduced in `1d60dafefeb8ea10b6cdb75258d829ff26191bb3`. Based on the registered tests, the demo app is able to perform the app switch to Venmo without it. Therefore, it does not appear to be required for Venmo.

PopupBridge/src/main/java/com/braintreepayments/api/internal/PopupBridgeJavascriptInterface.kt
pedrofsn: I think `isPayPalInstalled` would be better represented as `isPayPalAvailable`. However, this rename affects iOS and Web as well, so I recommend handling it as a follow-up patch in a separate PR.

pedrofsn: Nader is working on the web version. I recommend waiting to rename it until he returns on July 9, If we decide to do it. In another PR.

pedrofsn:
> I have the same concerns about this name as Jax. I think alignment is important but priority number one feels like a descriptive and easy to understand name

We can rename in a follow-up PR alongside iOS to reach cross-platform consensus on the exact name. Open to proposals here.

CC @jaxdesmarais @noguier

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeWebViewClient.kt
pedrofsn: Removed `setVenmoInstalled` from `onPageFinished`; `isVenmoInstalled` is already exposed via `@JavascriptInterface` getter

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
pedrofsn: https://github.com/braintree/popup-bridge-android/pull/109#issuecomment-4752443780

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
pedrofsn: Fixed analytics

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
pedrofsn: You're right, fixed.

pedrofsn:
> and, I assume, there is no other way to initialize a button?

Correct, @noguier.

PopupBridge/src/main/java/com/braintreepayments/api/internal/AppSwitchHandler.kt
pedrofsn: Venmo intercepts `account.venmo.com/braintree/checkout` via `shouldOverrideUriLoading`. PayPal uses `window.popupBridge.launchApp(url)` directly. The Venmo helpers (`isVennoAppSwitchuri`, `setVenmoInstalled`) mirror PayPal equivalents for this separate interception path.

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
pedrofsn: Sure

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeWebViewClient.kt
pedrofsn: `setVenmoInstalled` was removed entirely. `isVenmoinstalled` is now a computed property (`get() = context.isVenmoinstalled()`) in `PopupBridgeJavascriptInterface`, so it always reflects the current device state without needing to be called manually in `onPageFinished`.

noguier:
Hi there, before we proceed , can you please tell me if this behavior is expected ? Below happens when enablePayPalAppSwitch == false **or** when enablePayPalAppSwitch == true.  We do want this flow to work before merging. 

https://github.com/user-attachments/assets/aef6f8e9-3f62-47a4-b05d-606e86f39c07

PS: Venmo flow works great!

pedrofsn:
> Hi there, before we proceed , can you please tell me if this behavior is expected ? Below happens when enablePayPalAppSwitch == false **or** when enablePayPalAppSwitch == true.  We do want this flow to work before merging. 
> 
> https://github.com/user-attachments/assets/aef6f8e9-3f62-47a4-b05d-606e86f39c07
> 
> PS: Venmo flow works great!

Hi @noguier, thanks for flagging this and for the video. I looked into it in depth - here's what's happening.

**This flow was working and it should still work.** On 2026-06-15 I recorded the same Demo flow app-switching to the PayPal app in a sandbox environment (comment above). What you're seeing now is a regression, but it is **not** in this library / the Android client.

**Root cause: the URL the PayPal JS SDK generates when the yellow PayPal button is tapped.**
- Working before: `https://www.paypal.com/app-switch-checkout?appSwitchEligible=true&...&integration_artifact=PAYPAL_JS_SDK` -> app switch to the PayPal app.
- Now (sandbox): `https://www.sandbox.paypal.com/pay?...&native_xo=1&integration_artifact=PAYPAL_JS_SDK` (and `/checkoutnow?...&native_xo=1` for one-time flows) -> path is `/pay` (not `/app-switch-checkout`) and **`appSwitchEligible=true` is missing**, so it opens a Chrome Custom Tab and never app-switches.

The native side is healthy: `window.popupBridge.isPayPalInstalled()` returns `true` (verified live via CDP), the PayPal app is installed, and both the library and the PayPal app only recognize `/app-switch-checkout` as the app-switch deep link. Venmo works because it uses a different (WebView-intercept) mechanism, not this eligibility election.

**Proof it isn't the client code:** I rebuilt and ran three historical client trees on the same device today - `d61913d` (06-11), `0bf342f` (the exact tree from the 06-15 working video), and current `HEAD` - and **all three fail identically** (Chrome, `/pay?native_xo=1`). Code that demonstrably worked on 06-15 fails today, so the change is server/SDK-side. The timing also aligns with the popup bridge incident + backend changes around 06-22.

I filed a bug for the checkout/backend team to investigate why the sandbox session is no longer marked app-switch-eligible (no `appSwitchEligible=true` / `/app-switch-checkout`):
**DTINAPPXO-4878** - https://paypal.atlassian.net/browse/DTINAPPXO-4878

Happy to share the full captured URLs/logs or pair on it. Once backend confirms eligibility is restored (or if the app-switch URL contract changed), we can re-validate the flow here.

noguier:
Hi @pedrofsn , thank you for escalating this issue, feel free to post an update here, once it is fixed :)

jaxdesmarais:
Closing this while we wait on backend services to be re-deployed. Will reopen once ready for review/final testing.

PR #115, opened 2026-09-22
fix(security): Sanitize error message javascript.

mzlangreder:
### Summary of changes
- Harden error message handling by sanitizing the inputs. 
- Added tests to cover various sanitization scenarios.

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [ ] Claude 
- [x] Other (GLM, DeepSeek)

**How was AI used?**
Validate ticket, explain remediation, code the fix

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry
- [ ] Tested and confirmed payment flows affected by this change are functioning as expected

### Authors
- @mzlangreder


PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
```diff
@@ -245,3 +250,7 @@ class PopupBridgeClient @SuppressLint("SetJavaScriptEnabled") internal construct
                 "}")
     }
 }
+
+private fun String.escapeJsLineSeparators(): String {
+    return replace("\u2028", "\\u2028").replace("\u2029", "\\u2029")
```
saralvasquez: Just to check my knowledge, is it right that before adding this the line separators were escaped into character strings in the message?

noguier: just for the context why do we need this helper function, and why do we need to replace `u2028` with `u2029`?

PopupBridge/src/test/java/com/braintreepayments/api/PopupBridgeClientUnitTest.kt
```diff
@@ -235,15 +242,229 @@ class PopupBridgeClientUnitTest {
...
+        verify(exactly = 1) {
+            webViewMock.evaluateJavascript(withArg { javascriptString ->
+                assertTrue(javascriptString.contains("window.popupBridge.onCancel()"))
+            }, null)
+        }
+    }
+
+    @Test
+    fun `handleReturnToApp with hostile query params dispatches escaped success js`() = runTest {
+        val payload = "1); alert(document.cookie);//"
```
saralvasquez: Small nit: this uses the same hostile payload as the next test that is specifically testing for it. It might make the test a bit clearer if the payloads were different

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # PopupBridge Android Release Notes
 
+## unreleased
+
+* Harden JavaScript error delivery
+* The `err` argument of `window.popupBridge.onComplete` is now a JSON string on error paths
```
saralvasquez: Is this meant to be related to the first line? If so it might be better to indent it

saralvasquez (approved):
Couple of questions and nits but otherwise looks great! Thanks for adding such extensive and quality tests!

PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt
mzlangreder: U+2028/U+2029 are line/paragraph separators that are invisible, but are not stripped as part of the `JSONObject.toString()` function call so we added stripping for them here.

mzlangreder: Prior to the work in this PR, none of the content was being stripped so javascript could get interpolated into the string and executed if the `Failure` code path was exercised. Thankfully it's currently not exercised, but could be in the future so this is to protect against javascript inejection.

